PR #16, opened 2019-05-14
add alternative install location to install guide

zeue:
Without any extra configuration of the OS or Spotify, when using either the `spotify` or `spotify-dev` packages from the AUR, my Spotify was located in `/opt/spotify` and not `/usr/bin/spotify`. Might help someone else in the future to have this noted in the README file.


khanhas:
Thank you!
I changed and left a note in there. 
I also moved it to wiki page.

Issue #12, opened 2019-04-11
Use ~/.config/spicetify for configuration instead of ~/.spicetify


nsilvestri:
It's common practice to have user configuration files for Linux applications in `~/.config` (for example: pcmanfm, polybar, GTK, Qt, redshift, rofi, ranger, xfce). There's no reason to have spicetify use its own `~/.spicetify` directory, as it's another hidden file cluttering up `~`. I suggest that Spicetify start using `~/.config/spicetify` to conform with configuration file location convention.

khanhas:
That makes sense. I will change in next version.
Thank you!

vesim987:
To be more precise you should first try to read `XDG_CONFIG_HOME` environment variable and fallback to `~/.config/` when the env var is missing.

khanhas:
I added this feature in v0.8
Please check it out and tell me if it works correctly.

zeue:
working for me

zeue:
@khanhas works for me! cheers

Dorsug:
It is also common practice on macos. Moreover `~/spicetify_data` isn't a hidden file.
I'm completly happy for now using `SPICETIFY_CONFIG="$XDG_CONFIG_HOME/spicetify`, but is there any reason why it isn't the same on macos?

jonathanlaniado:
> It is also common practice on macos. Moreover `~/spicetify_data` isn't a hidden file.
> I'm completly happy for now using `SPICETIFY_CONFIG="$XDG_CONFIG_HOME/spicetify`, but is there any reason why it isn't the same on macos?

+1, is this something macOS users can also do?

khanhas:
@jonathanlaniado Have you tried it? I added that feature ages ago.

jonathanlaniado:
Yes, thanks for checking. Was able to make it work with $HOME. :)

alichtman:
This is broken for me on `spicetify v1.0.2`

khanhas:
@alichtman XDG is only for linux
In all platform, use `$SPICETIFY_CONFIG`

alichtman:
Thanks.

Why can't the XDG spec be supported on macOS? It is UNIX compliant.

itsmeboy0:
> Thanks.
> 
> Why isnt't the XDG spec supported on macOS? It is UNIX compliant.

+1 for this. Not just for macOS, this should be supported even on Windows. I do not like clutter on **any platform** at all.

alichtman:
[...]

PR #19, opened 2019-05-23
update README.md to reflect Linux using $XDG_CONFIG_HOME for configuration files

nsilvestri:
With the change to use the `$XDG_CONFIG_HOME` variable for configuration files in #12, the readme needed to be updated. I've opted to hard code in `~/.config/spicetify` rather than using `$XDG_CONFIG_HOME/spicetify` since `$XDG_CONFIG_HOME` is probably not set as most users don't change the default. Maybe there's a smarter way to indicate that this application is XDG Base Directory compliant?


khanhas:
Thank you. I fixed and moved it to wiki page

PR #37, opened 2019-07-11
Update WebNowPlaying extension to send track position with milliseconds

GrumpyBear57:
Dependent on WebNowPlaying.dll v0.4.1 (tjhrulz/WebNowPlaying#32)

### DIFF
--- a/Extensions/webnowplaying.js
+++ b/Extensions/webnowplaying.js
@@ -149,6 +149,12 @@
         const seconds = Math.floor(timeInMs / 1000);
         const minutes = Math.floor(seconds / 60);
 
-        return `${Math.floor(minutes / 60)}:${minutes % 60}:${seconds % 60}.${pad(timeInMs % 1000, 3)}`;
+        const padSec = pad(seconds % 60, 2);
+        const padMs = pad(timeInMs % 1000, 3);
+
+        if (minutes < 60) {
+            return `${minutes}:${padSec}.${padMs}`;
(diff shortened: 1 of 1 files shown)


khanhas:
Hi! Thanks for the pull request.
As soon as your commit on WebNowPlaying plugin is merged, I will merge this.

One thing though, instead of using condition in https://github.com/khanhas/spicetify-cli/pull/37/commits/08c7de179f57c8415aa8ff8d749c749bcd6042fd#diff-a3d829c84471468090e73272c58850b1R57
Just add field `POSITIONMS` in https://github.com/khanhas/spicetify-cli/pull/37/commits/08c7de179f57c8415aa8ff8d749c749bcd6042fd#diff-a3d829c84471468090e73272c58850b1R24

Also, why you remove zero-padding in `convertTimeToString`?

GrumpyBear57:
You're right, totally missed the fact that I didn't need the conditional check anymore. Originally when I was working on WNP.dll I did need it to send both POSITION and POSITIONMS, but don't anymore since everything that POSITIONMS on WNP.dll gets mirrored onto POSITION.

I removed the zero padding for everything but ms because in WNP it now uses TimeSpan.Parse() instead of custom math that is dependent on exact formatting. So it sees the new 1:2:3 the same as 01:02:03. The only thing that makes a difference now is milliseconds, since 1.1 seconds is much different than 1.001 seconds.

PR #97, opened 2019-10-27
Update install.ps1

LivingWithHippos:
Adding the parameter -UseBasicParsing to Invoke-WebRequest fixes the error appearing when
Internet Explorer engine is not available or its first-launch configuration is not complete.


khanhas:
Thank you!
That's clever.

PR #160, opened 2020-03-26
Added prev and next track, scroll faster with U and D

mackarelfish:
Added a few shortcuts :+1:


khanhas:
Thank you for contribution.
I already made the extension customizable enough for user to change on their own. And I don't think these new additions are necessary.

PR #260, opened 2020-08-20
fullAppDisplay now with album release date and optional icons

JulienMaille:


khanhas:
I'm not approved commit 2ff9e77, please revert back.

JulienMaille:
> I don't approved commit [2ff9e77](https://github.com/khanhas/spicetify-cli/commit/2ff9e7797a270ab74e91836696805479e2b04999), please revert back.

Ok, but have you noted that the existing button for the similar fullscreen is like this
image

khanhas:
Yup, that's the exact reason. FullAppDisplay is not full screen function and I don't want to these 2 buttons having same icon.

PR #272, opened 2020-08-30
newReleases.js: Show Podcasts only and Ignore All

yougotwill:
First thank you for making this amazing program!

This is my first contribution to this project so I'm sorry if I've made any mistakes.

This PR allows the newReleases.js extension to show Podcasts only. I didn't want all the new releases information because I use the Release Radar every week but I thought that having new podcast information could be useful!

I have also added an Ignore All option to clear all new release notifications.

Let me know if there is anything that needs to be fixed.

### DIFF
--- a/Extensions/newRelease.js
+++ b/Extensions/newRelease.js
@@ -340,7 +340,7 @@
 
     // Add context menu items for Notification button
     const checkURI = ([uri]) => uri === "spotify:special:new-release"
-    new ContextMenu.Item(
+    const podcastContextMenuItem = new ContextMenu.Item(
         "Podcast",
         function () {
             const state = !BUTTON.isFetchingPodcast()
@@ -350,7 +350,8 @@
         },
         checkURI,
         BUTTON.isFetchingPodcast() ? "check" : null,
-    ).register()
+    )
+    podcastContextMenuItem.register()
 
     new ContextMenu.Item(
         "Followed artists only",
@@ -382,6 +383,10 @@
             const state = !BUTTON.isPodcastOnly()
             BUTTON.setPodcastOnly(state)
             this.icon = state ? "check" : null
+            if (state && !BUTTON.isFetchingPodcast()) {
+                BUTTON.setFetchingPodcast(true)
+                podcastContextMenuItem.icon = "check"
+            }
             main()
         },
         checkURI,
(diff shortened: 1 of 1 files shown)


khanhas:
Thanks you for your first contribution to the repo, it works great!
One thing tho, I think `Podcast Only` option should enable `Podcast` option too (if `Podcast` isn't already enabled). What do you think?

yougotwill:
Wow that was fast! I think that is a good idea. I'll add it now. I've just added a "Clear all" button to the context menu on my local branch. Should i rename it to "Ignore all" or is Clear ok?

khanhas:
I think `Ignore all` fits better.

yougotwill:
So I've looked into updating the Podcast ContextMenu Item when you click the "Podcast Only" button but I have one issue.

With the way the ContextMenu.Item is built in the spicetifyWrapper there is not way to update the icon outside of its own context (i.e when you use new ContextMenu.Item).

So far I update isFetchingPodcast correctly when the "Podcast only" button is clicked but I cannot get the icon to update without clicking on the "Podcast" button itself.

Do you have any advice? Sorry for the noobish question

khanhas:
Store Podcast context menu item in a variable:
```js
const podcastContextMenuItem = new ContextMenu.Item("Podcast", ...)
podcastContextMenuItem.register()
```
Then just use it in Podcast Only callback to change its icon:
```js
new ContextMenu.Item(
        "Podcasts only",
        function () {
            const state = !BUTTON.isPodcastOnly()
            BUTTON.setPodcastOnly(state)
            this.icon = state ? "check" : null
            if (state && !BUTTON.isFetchingPodcast()) {
                BUTTON.setFetchingPodcast(true)
                podcastContextMenuItem.icon = "check"
            }
            main()
            ...
```

yougotwill:
Ah ha! I was so close. I didn't move register() to be called on the const variable I had made! Thank you so much was a good learning experience. Tested and everything seems to work will push now and update the PR to include information for 'Ignore all'.

yougotwill:
ready for merging!

khanhas:
Ignore All implementation is wrong, it only removes every items in the list, not actually ignore any of them. If you reload the client, items come back.

PR #293, opened 2020-09-19
Open window in fullscreen in fullAppDisplay.js

diegodalbosco:
I think that the "full app display" extension should open in fullscreen. I made two little changes to request and exit fullscreen, I saw in the zlink.bundle.js file that these are the methods used by the Spotify fullscreen functionality.


khanhas:
Thank for your contribution, it would be a nice addition.
But I don't want it to be fullscreen by default. Could you make it as a Context Menu option?

diegodalbosco:
> Thank for your contribution, it would be a nice addition.
> But I don't want it to be fullscreen by default. Could you make it as a Context Menu option?

@khanhas I added the context menu option for full screen

PR #300, opened 2020-09-26
XDG base spec compliance on OSX

Starchery:
A lot of people (myself included) found it strange that the spicetify folder is unhidden in `$HOME` on OSX: `~/spicetify_data`

It's easy to fix it yourself by doing 
```sh
$ export SPICETIFY_CONFIG="$HOME/.config/spicetify"
```
but it should conform to the [XDG base dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) by default, like it does on Linux. It's the closest thing to a cross-platform standard we have.

Alternatively, it could default to something like `$HOME/.spicetify`, as it does right now on Windows. But if we're changing the directory, we might as well make it standards compliant--most modern applications respect `$XDG_CONFIG_HOME`. Changing the Windows configuration is outside the scope of this PR.

### TODO
The basic functionality is there--the obvious caveat is backwards compatibility. I propose that we check if `~/spicetify_data` exists if `runtime.GOOS == "darwin"`, and using that as the folder if it does, falling back to 
```sh
"${XDG_CONFIG_HOME:-"$HOME/.config"}/spicetify"
``` 
otherwise. The user can move the location whenever they want, and backwards compatibility is saved. This isn't the only option though, and I'd love to hear feedback.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -10,5 +10,6 @@ golang.org/x/net v0.0.0-20200506145744-7e3656a0809f/go.mod h1:qpuaurCH72eLCgpAm/
 golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20200116001909-b77594299b42/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd h1:xhmwyvizuTgC2qz7ZlMluP20uW+C3Rm0FD/WLDX8884=
 golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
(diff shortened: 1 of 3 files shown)


x3ro:
This worked perfectly on my machine and would be much appreciated 👍🏼 Also rebases to the current master branch without conflicts.

khanhas:
Thank you for Starchery contribution and @x3ro for testing it out.
But it will break for MacOS users who use default `spicetify_data` location. I will change it myself and it will be available in next version.

strafe:
> Thank you for Starchery contribution and @x3ro for testing it out.
> But it will break for MacOS users who use default `spicetify_data` location. I will change it myself and it will be available in next version.

@khanhas there's been 2 releases since this, has this been added yet?

khanhas:
Have you checked?

strafe:
> Have you checked?

It wasn't in the release notes for either of the updates.

khanhas:
I mean have you checked if it works with env XDG_CONFIG_HOME?

strafe:
> I mean have you checked if it works with env XDG_CONFIG_HOME?

Not as of yet because I have `SPICETIFY_CONFIG` set.

strafe:
@khanhas I see there is now some logic related to it in `cmd.go` but it's still not compliant. The spec states:
> If $XDG_CONFIG_HOME is either not set or empty, a default equal to  $HOME/.config should be used.

Currently, `$HOME` is used instead of `$HOME/.config` (like Linux).

khanhas:
Sorry, I won't move default location for MacOS. Please just use $XDG_CONFIG_HOME or SPICETIFY_HOME if you want it locates elsewhere.

strafe:
> Sorry, I won't move default location for MacOS. Please just use $XDG_CONFIG_HOME or SPICETIFY_HOME if you want it locates elsewhere.

Why is that? I figure it should always check `$HOME/spicetify_data` if they are an existing user, but new installs that lack that folder should move to the new location? It should be possible to be backwards compatible.

itsmeboy0:
Any updates on this?

PR #307, opened 2020-10-12
Allow toggling of Full App Display with the F11 key.

mendelsimon:
Allow the use of the F11 key to toggle the Full App Display.


khanhas:
Thanks for your contribution! I think it's a great idea.

Issue #262, opened 2020-08-21
getAblumArtColors() returns invalid colors


JulienMaille:
I know this is not your code but 
```javascript
Spicetify.getAblumArtColors("spotify:album:3w26tAxCUFYYzeFMq5qNJg")
        .then((colors) => {
            console.log(colors)
        })
```
`{VIBRANT: "#459a6", PROMINENT: "#e9d1d3", DESATURATED: "#459a6", LIGHT_VIBRANT: "#459a6", DARK_VIBRANT: "#459a6", …}`

With only 5 characters, #459a6 is not a valid hex color.

Looking at the json returned by spotify this is what we get
```
color_swatches: Array(6)
0: {preset: "VIBRANT", color: 285094}
1: {preset: "PROMINENT", color: 15323603}
2: {preset: "DESATURATED", color: 285094}
3: {preset: "LIGHT_VIBRANT", color: 285094}
4: {preset: "DARK_VIBRANT", color: 285094}
5: {preset: "VIBRANT_NON_ALARMING", color: 285094}
```

Why do you pad to 3 and not to 6 here:
https://github.com/khanhas/spicetify-cli/blob/70aecd2fa1d284f4c9e94d6c1479c6cf0a18426e/jsHelper/spicetifyWrapper.js#L1392

BTW, there is a typo in `getAblumArtColors` should be `getAlbumArtColors` :-)
BTW2, spotify `colorextractor` is pretty bad, almost all returned colors are identical

PR #315, opened 2020-10-28
Fix typos

Zedeldi:
Hi there,

As #262 mentioned, there's a small typo in `getAblumArtColors()`. It should be `getAlbumArtColors()`.
This is referenced in both `globals.d.ts` and `jsHelper/spicetifyWrapper.js`.

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -210,7 +210,7 @@ declare namespace Spicetify {
      * @param uri is optional. Leave it blank to get currrent track
      * or specify another track uri.
      */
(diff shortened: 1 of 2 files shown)


khanhas:
Don't worry about it. `getAblumArtColors` will be deprecated soon. In replace, `Spicetify.colorExtractor` is more versatile and favourable.

JulienMaille:
@khanhas cool will try colorExtractor

khanhas:
Well, I don't think I will include any function that is not provided by Spotify API.

PR #323, opened 2020-11-05
fullAppDisplay.js: Fix song duration after skipping

### LINKED ISSUE #312: [Extensions] Duration shift with fullAppDisplay.js
First of all, thank you for the great software!

### Context
When using the `fullAppDisplay.js` extension.

### Issue
To reproduce:
1. toggle the `fullAppDisplay` mode `on` by clicking on the icon while playing a song in a playlist or in an album
2. click on `>` to skip to next song
> The song duration doesn't change. 
3. click again on `>`
> The son duration changed but it is the duration of the previous song in the playlist/album

### Demonstration
issue_fullAppDisplay

In the demo above, the first album durations are `3:59`, `4:35`, `3:47` and `2:45`. But with the first duration being repeated, we have `3:59`, `3:59`, `4:35`, `3:47`.

### Spicetify version
I'm using `spicetify v1.1.0`, installed with the standard PowerShell procedure on Windows 10.

Please let me know if you need any additional information!

LambArchie:
Fixes #312
Found that Spicetify.Player.getDuration() doesn't immediately update to the new song length after changing track so added a delay in

Delay could be lowered by more than 20ms but wasn't sure how long it needs to be delayed for, so the getDuration performs properly

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -319,10 +319,8 @@ body.fad-activated #full-app-display {
             album.innerText = Spicetify.Player.data.track.metadata.album_title + " • " + album_date
         }
         if (CONFIG.enableProgress) {
-            // Delayed as getDuration doesn't immediately update
-            setTimeout(function (){
-                durr.innerText = Spicetify.Player.formatTime(Spicetify.Player.getDuration())
-            }, 20);
+            // Not using Spicetify.Player.getDuration() due to bug
+            durr.innerText = Spicetify.Player.formatTime(Spicetify.Player.data.track.metadata.duration)
         }
     }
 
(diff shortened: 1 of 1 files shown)


LambArchie:
@khanhas Any chance of this being merged soon? Or does something need changing?

khanhas:
Thanks for your contribution
I don't actually like the setTimeout solution, so I tried a few to find what's best.
Can you help me test with this one
```
durr.innerText = Spicetify.Player.formatTime(Spicetify.Player.data.track.metadata.duration)
```

LambArchie:
Tested the code you provided and couldn't find any issues with it, and yeah that is a much better solution. Going to update the patch to the code you provided.

PR #343, opened 2020-12-01
Add option to show all artists in fullAppDisplay.js

OlegWock:
Looks like this 
image

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -301,12 +301,19 @@ body.fad-activated #full-app-display {
                 .trim()
         }
         title.innerText = rawTitle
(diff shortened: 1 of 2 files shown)


OlegWock:
Oh, sorry, it seems I added another fix to the same PR instead of a new one (I'm still not very good at dealing with PRs here 😅), I hope it's okay to have two fixes in one PR

Quick description of fix:
When fetching new releases, podcast episodes have id `spotify:episode:smth` (which is correct). But when listening to a podcast, `Player.data.track.metadata.context_uri` will return `spotify:show:smth` so the extension will save it as listened (which is incorrect) thus episode will be never marked as listened. `Player.data.track.uri` contains the correct id so for podcasts we're using it

PR #367, opened 2020-12-31
shuffle+: Add support for nested folders

iurimateus:
Currently, 'Play with Shuffle+' is broken for folders that contain other folders or reside within another.

Returning `TypeError: Cannot read property 'map' of undefined` or `TypeError: Cannot read property 'rows' of undefined`, in the second case.

This PR fixes it by applying a recursive search.

### DIFF
--- a/Extensions/shuffle+.js
+++ b/Extensions/shuffle+.js
@@ -12,11 +12,11 @@
         return;
     }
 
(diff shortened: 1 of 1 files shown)


khanhas:
Works very well, thank you!

PR #400, opened 2021-02-05
Extension showsNewsRemover.js

georgwacker:
Removes shows/news sections from Home


khanhas:
Extensions I created and packed are more of a stash of thing I personally use. And they serve as examples on how to make an extension and use Spicetify API. I rarely accept any extension from contributor since this repo does not a function as package manager repo.
Hope you understand.
You can always put it up on your personal github, tag it with `spicetify` and add a descriptive readme. I'm sure it will gain visibility soon. One more thing, your filter code in your extension only targets English language. It definitely wouldn't work on others.

PR #460, opened 2021-03-19
fix ContextMenu deregister function, Set does not have a remove method

nmattis:
Noticed while writing an extension that needed to deregister a created context menu item in some situations that the function was using a .remove method which does not exist on a Set object. This change switches it to the delete method so that deregister does not error now.

Console error when attempting to deregister on version 1.2.1 for reference:
```
TypeError: itemList.remove is not a function
    at Item.deregister (spicetifyWrapper.js:1527)
    at extension.js:126
```


khanhas:
Thank you!

PR #462, opened 2021-03-20
Create playlistOverlap extension

reteps:
image

The idea of this extension is to show the songs that overlap between all your playlists as a tooltip. All loading is done using async, so the page does not load any slower until the final result of overlap is calculated.

Let me know if you have any code or other questions, this code isn't the prettiest.

Note that caching of results is not done at all. This was to avoid the issue of adding a song to a playlist, then the computed overlap would be messed up.

### DIFF
--- a/Extensions/playlistOverlap.js
+++ b/Extensions/playlistOverlap.js
@@ -1,0 +1,199 @@
+/*
+Peter Stenger
+3/2/2021
+
+
+Window Event Listener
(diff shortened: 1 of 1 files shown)


khanhas:
Thank you for contribution but I'm sorry, I personally don't accept new extension pull request on this repo.
You might want to find an spicetify extensions repo running by community to put this up.

PR #510, opened 2021-04-12
Fade album art and background image in fullAppDisplay extension

julthon:
I added smooth fading for the fullAppDisplay extension since a song change flickered for a short period while loading the image. This is done by preloading the new image and updating the actual image source 0.5s later as well as adding a background transition.

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -47,6 +47,7 @@
     padding-bottom: 100%;
     border-radius: 15px;
     background-size: cover;
+    transition: background-image 1s ease-in-out;
 }
 #fad-art-inner {
     position: absolute;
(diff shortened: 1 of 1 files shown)


khanhas:

The transition lags a lot when Spotify window is maximised or in full screen mode. Does it happen to you?

julthon:
The transition is smooth on my workstation pc, but not so much on my low-end notebook. I think it's still way better than the sudden change without transition.
A possibility would be to include a context menu option to toggle the transition animation.

khanhas:
Adding context menu option would be wise, thank you.

khanhas:
Hi, I just made a few modification on UI update code. Now it's event based instead of waiting 500ms. Can you test it out on your end and see if the animation is still smooth:
https://raw.githubusercontent.com/khanhas/spicetify-cli/master/Extensions/fullAppDisplay.js

julthon:
Hi, the event based trigger seems to work just as well! I added another css property which drastically reduces the lag during the animation. I will open another pull request.

PR #526, opened 2021-04-20
Update fullAppDisplay.js

McStecca:
Changed the annoying minimize button to access fullscreen to the maximize one

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -396,7 +396,7 @@ body.fad-activated #full-app-display {
 
     // Add activator on top bar
     const button = document.createElement("button")
-    button.classList.add("button", "spoticon-minimise-16", "fad-button")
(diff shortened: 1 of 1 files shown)


JulienMaille:
Have a look here :)
https://github.com/khanhas/spicetify-cli/pull/260#issuecomment-678099470

McStecca:
It doesn't make sense, i understand that fullAppDisplay.js and fullscreen are two different things with different functions, but the minimize button is unaesthetic. I could search for a different icon if it's necessary

JulienMaille:
I agree the minimise icon is misleading, you can have a look at availables icons here: https://github.com/morpheusthewhite/spicetify-themes/pull/272#issuecomment-779090321
maybe the "tv" icon?
image

khanhas:
If it's unaesthetic to you, just change it locally whatever you prefer ;)

PR #538, opened 2021-04-23
Requesting to add to the collection of extensions

lucasburlingham:
Submitting extension to remove the ad banner at the bottom of the client's main section. 

I block ads by default with PiHole, leaving a large black box. I decided to try to not render the box with a CSS `display: none !important;`, but that didn't work because the CSS is being loaded first, not the JS which most likely inserts the ads dynamically. Thus, I figured that since the JS from the extensions get loaded after the `init.js` file (which renders the Spotify client window and all elements contained therein), an extension written in JS would work best.

 It does work, but may not be well-received. My argument for using this is that the advertisement is being loaded, but not displayed. The user can disable the extension at any time.

Thanks for any consideration the maintainers of this _awesome_ project give!

(BTW, I have been shipping a modified version of Ubuntu Desktop with this project included: [JovarkOS on SourceForge](https://sourceforge.net/projects/jovarkos/). Thanks for the hard work!)

### DIFF
--- a/Extensions/removeAds.js
+++ b/Extensions/removeAds.js
@@ -1,0 +1,4 @@
+document.onload = function removeAd {
+	var element = document.getElementById("view-leaderboard-ad");
+	element.parentNode.removeChild(element);
+}


khanhas:
Thank you for your contribution, but spicetify is not for ad blocking purpose and never support ad-blocking in any way.

PR #542, opened 2021-04-26
Disable backface-visibility in fullAppDisplay extension

julthon:
This change improves the animation performance when fading between songs.


khanhas:
How the hell would adding more blur makes it so much smoother?
Anyway, thank you for these changes.

PR #558, opened 2021-05-08
Improve shuffle+.js

SeitaHigashi:
Add "Add to Queue and Shuffle+" function.
Useful for shuffling multiple playlists and albums.
I created it for myself, but I hope you can integrate it.

### DIFF
--- a/Extensions/shuffle+.js
+++ b/Extensions/shuffle+.js
@@ -89,6 +89,26 @@
         if (headers.length > 0) clearInterval(iframeInterval);
     }, 500)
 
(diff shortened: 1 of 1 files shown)


khanhas:
Good idea.
Would mind move the URIs checking function to separated function and reference it in both context menus initializer? Instead of having 2 exact copies like they currently are.

SeitaHigashi:
OK. Should I do it?

khanhas:
Yeah you should.
Wonder why you ask that.

PR #657, opened 2021-05-22
FullAppDisplay : Prevent animation when reopening extension with fade enabled

BluTree:
Fix a little bug when the animations are enabled that makes the app to animate between the last displayed song and the current song when we reopen the extension.

The fix for the front cover animation is clearly a workaround. I don't have any other idea as I don't have enough knowledge in web in general.


khanhas:
I can't see why it would be a problem.
Plus, you edited bunch of other unrelated stuff, I can't accept this pull request.

PR #664, opened 2021-05-23
Fix NewReleases For Followed Artists

Exhigh:
This is a rough draft for the fix


khanhas:
No longer relevant. Artists list request only returns followed artists now.

PR #701, opened 2021-05-30
Better Dracula colors

Akasiek:
This new Dracula color bugged me for a bit and I want to change it to the more classic colors.
With the changes (beside the long volume bar): 
Screenshot_1


khanhas:
Thanks for your contribution. But I like yellow better.

PR #706, opened 2021-05-30
IMP: thinner icons for lyrics-plus and new-releases

JulienMaille:
FIX: https://github.com/khanhas/spicetify-cli/issues/705


khanhas:
Thank you!

PR #754, opened 2021-06-09
Extra Dracula color

Akasiek:
Add extra color for Dracula variant. More Dracul'ish...

Image for reference:
Screenshot_1

Maybe this method will work? It worked in Dribbblish...


afonsojramos:
This makes more sense in spicetify-themes or to just leave it as a personal theme, but thanks for your PR!

PR #775, opened 2021-06-16
Moved Legacy Instructions to the Wiki

afonsojramos:
Updated Wiki with the legacy installation instructions ([Installation](https://github.com/khanhas/spicetify-cli/wiki/Installation)).


elijaholmos:
why is this necessary? There are still several users that prefer the legacy UI to the new one.

afonsojramos:
> why is this necessary? There are still several users that prefer the legacy UI to the new one.

I'm not deleting this from existence, I'm just moving it to the Wiki, just like the other instalation scripts, to not confuse new users, which has been happening a lot if you take a look at the issues that have been being opened.

elijaholmos:
@afonsojramos okay, I see what you mean. I misunderstood the original PR, and I have seen some issues that you mentioned where users skimmed the readme and installed legacy Spicetify by mistake. Although this is user error, I think it would clean up the repo more if all the instructions were in one place 👍

PR #858, opened 2021-07-09
Fix:Close popup when click is Outside

daksh2k:
Closes the Config popup when click is outside of the popup.

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -909,6 +909,14 @@ class _HTMLGenericModal extends HTMLElement {
         this.querySelector("button").onclick = this.hide.bind(this);
         const main = this.querySelector("main");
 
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):
Looks good to me!

Works as intended.

jsHelper/spicetifyWrapper.js
```diff
@@ -909,6 +909,14 @@ class _HTMLGenericModal extends HTMLElement {
         this.querySelector("button").onclick = this.hide.bind(this);
         const main = this.querySelector("main");
 
+        const toCheck = this.querySelector('.GenericModal');
```
afonsojramos, line 912: Can you remove this `const` declaration since it's only used once?

daksh2k: Sure :)
I have added another commit.

PR #1016, opened 2021-08-31
Patch css-map

CharlieS1103:
This will hopefully restore the functionality of some custom apps


src/preprocess/preprocess.go
```diff
@@ -274,7 +273,7 @@ func exposeAPIs_main(input string) string {
 	reAllAPIPromises := regexp.MustCompile(`return (\w+=\w+\.sent),\w+\.next=\d,(Promise.all\(\[(\w\.getSession\(\),)([\w\(\)\.,]+?)\]\))([;,])`)
 	allAPIPromises := reAllAPIPromises.FindAllStringSubmatch(input, -1)
 	for _, found := range allAPIPromises {
-		splitted := strings.Split(found[3] + found[4], ",")
```
itsmeow: Don't commit formatting changes like this, you should remove both of these commits from the tree and force push. Otherwise you are just messying up the commit history with changes that do nothing

itsmeow (approved):

PR #1026, opened 2021-09-02
Add minor version tag support for CSS mappings

itsmeow:
Will now only pull from highest minor version tag (ex: 2.6.x) instead of master branch, allowing for moderate backwards compatibility. This change also makes the development builds (ones without a version tag) use local CSS mappings only, for convenience.


afonsojramos (approved):
Looking good! From 2.6.2 CSS Maps will be updated more easily 👌

PR #1034, opened 2021-09-04
Update community badge

morpheusthewhite:


itsmeow:
Might I suggest we add the Discord to the README as well?

morpheusthewhite:
Do we want to? Isn't it better to group the community around a single place instead of spreading them across different platforms?

CharlieS1103:
@morpheusthewhite As far as I can tel the discord is much more active than the reddit, and I find it to be a much easier platform to use when helping people with spicetify or working with people to help fix spicetify. Especially with the new threading function.

TheGeeKing:
Yep, totally agree with Charlie.

Discord is the best place to discuss and help about Spicetify, you should check it a lot of people are active on it. It surprises me that the discord wasn't easily accessible to everyone on the github. We should definitely add it to the front page.

morpheusthewhite:
Ok then, I'll add it.

Thanks for the feedback btw.

TheGeeKing:
You're welcome. https://discord.gg/VnevqPp2Rr

morpheusthewhite:
I asked the discord admin to enable the badge on the server, it will soon be shown properly

afonsojramos (approved):

PR #1071, opened 2021-09-19
Fix: Popup hiding behind UI without closing

daksh2k:
Sometimes the popup hides behind UI after changing something. This fixes that.


afonsojramos (approved):
I mean, `z=10` would probably suffice, but since this is a modal either way it doesn't make a difference.

Thank you!

Issue #1090, opened 2021-09-23
Synced Lyrics no do not work after playing song with no lyrics


znicuuu:
## ℹ Computer information

- Spotify version: 1.1.68.632.g2b11de83
- Spicetify version: 2.6.4

### ✔️ Expected result
popupLyrics showing lyrics normally

### ❌ Actual result
popupLyrics saying theres no lyrics for every song

## 📷 Screenshots
![image](https://user-images.githubusercontent.com/73033465/134532636-4dcbec71-e7e7-4b19-8e03-6ef0d0d82498.png)

itsmeow:
Can you see if 2.6.5 fixes this for you?

afonsojramos:
Please run `spicetify upgrade` and then `spicetify restore backup apply` to check if it fixes it.

Lseoksee:
> 실행 `spicetify upgrade`한 다음 `spicetify restore backup apply`해결되는지 확인하십시오.
> 그러나 실제로 먼저 더 알려진 음악으로 테스트할 수 있습니까? 나는 [Planet B - King Gizzard와 Lizard Wizzard를](https://open.spotify.com/track/3YCiB91fQ6aGH3KkBEyriG?si=e32336932cea4092) 제안할 수 있습니다.[](https://open.spotify.com/track/3YCiB91fQ6aGH3KkBEyriG?si=e32336932cea4092)

If you play some songs without lyrics and play songs with lyrics again, it says there are no lyrics. This is the same in lyrics-plus.

Lseoksee:
> Please run `spicetify upgrade` and then `spicetify restore backup apply` to check if it fixes it.
> But actually, first can you test with a more known music? I can suggest [Planet B - King Gizzard and the Lizard Wizzard](https://open.spotify.com/track/3YCiB91fQ6aGH3KkBEyriG?si=e32336932cea4092)

If you play some songs without lyrics and play songs with lyrics again, it says there are no lyrics. This is the same in lyrics-plus.

afonsojramos:
@Lseoksee tested it myself, you are right! I'll investigate further!

itsmeow:
I cannot reproduce this myself. Playing a song with no lyrics then playing another, it still works. On 2.6.6.

AntServiceApps:
[...]

PR #1101, opened 2021-09-26
Fix Lyrics Plus Karaoke position error

### LINKED ISSUE #1099: The lyrics-plus custom app isn't working as intended (stuck at the initial state)
## ℹ Computer information

- Spotify version: 1.1.68.632-1
- Spicetify version: 2.6.6

## 📝 Provide detailed reproduction steps (if any)

1. Download Spicetify and enable the custom app lyrics-plus
2. Play a song synced on Musixmatch
3. Play a moment when a line is being singed

### ✔️ Expected result

_What is the expected result of the above steps?_

The vocal line should be focused and the visuals have to update.

### ❌ Actual result

_What is the actual result of the above steps?_

Nothing happens.

## 📷 Screenshots

_Are there any useful screenshots?_

You can see in the following screenshot that I'm playing a song and I'm at the near-end of the track, where lyrics has almost ended, however, lyrics-plus still shows the start of the lyrics, nothing has moved from the initial state.

## 🔎 Research trail

You might notice, in the previous screenshot, that I have my element inspector opened, and the console consistently spits "Cannot read property 'position_as_of_timestamp' of undefined" errors happening on [this line](https://github.com/khanhas/spicetify-cli/blob/master/CustomApps/lyrics-plus/Pages.js#L102). That would mean that PlayerState's value of [position_as_of_timestamp](https://github.com/khanhas/spicetify-cli/blob/master/globals.d.ts#L42) is never written, and, lyrics-plus using this value to update, would ultimately crash too.

When fidgeting with the Console tab, I noticed `Spicetify.Player.origin2.state.playback` returned `undefined`. The problem probably comes from the fact that this value does not exist. As I'm fairly inexperienced with Typescript, I don't really know where to look for to repair such issue. I'm not sure if that's caused by a Spotify update, for instance if they changed the name of their variables (if the value of `state` or `playback` is provided by Spotify themselves) or if the issue comes from spicetify not initializing `playback` correctly for some reason. I tried checking for files containing of `playback` in spicetify but I couldn't find anything interesting.

itsmeow:
Fixes #1099

### DIFF
--- a/CustomApps/lyrics-plus/Pages.js
+++ b/CustomApps/lyrics-plus/Pages.js
@@ -98,8 +98,7 @@ const KaraokeLyricsPage = react.memo(
         const [position, setPosition] = useState(0);
         useTrackPosition(() => {
             if (!Player.data.is_paused) {
-                const playback = Spicetify.Player.origin2.state.playback;
-                setPosition(playback.position_as_of_timestamp + (Date.now() - playback.timestamp));
+                setPosition(Spicetify.Player.getProgress());
             }
         });
 


Ballasi:
Works for me!

afonsojramos:
I'm not sure if this is fixing anything as everything detailed in #1090 is still happening.

@itsmeow do you have a MusicXMatch token?

afonsojramos:
Sometimes I wonder if it is just lack of lyrics availability from MusicXMatch 🤔

itsmeow:
This PR does not attempt to address #1090, this only addresses the error in #1099. The two errors are completely unrelated and in different apps.

PR #1154, opened 2021-10-11
Add manifest.json for spicetify marketplace

TheGeeKing:
Add manifest.json with v2 spicetify extensions


CharlieS1103:
@afonsojramos It's certainly not a problem, we spoke about it in the disc. They just said it was unconventional (Which is fine since it works perfectly for what we're doing.)

PR #1159, opened 2021-10-12
Add action to automatically update AUR and homebrew on tag creation

itsmeow:
Whenever a tag/release is created, it makes a request to my server which then pushes an update to AUR and homebrew automatically.

Unfortunately, I have to host the application on my server because we don't have access to repository secrets. Otherwise, I would just write a general github action with Docker or something and pass the SSH key in the secrets.

Source of the server application is here: https://github.com/itsmeow/spicetify-cli-update/

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,16 @@
+name: Update Homebrew/AUR from tag
+
+on:
+    push:
+        tags:
+            - "v*"
+
+jobs:
+    update:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Request
+              uses: fjogeleit/http-request-action@master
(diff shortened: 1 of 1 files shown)


afonsojramos:
I feel a bit reluctant with using your server in particular as it is somewhat hard to guarantee its availability. What do you think of hosting this in Vercel (free hosting) for example? This way the deployment could be better managed by more than one person.

itsmeow:
Also, it has to be hosted on my hardware, it uses my SSH keys, which I'm not willing to put elsewhere. I'm the only one with access to the AUR repo.

afonsojramos:
@itsmeow Vercel supports secrets though.

PR #1179, opened 2021-10-18
shuffle+: Fix failed request and exception on empty folders.

### LINKED ISSUE #1152: Shuffle+ Not Working
## ℹ Computer information

- Spotify version: 1.1.69.612.gb7409abc
- Spicetify version: 2.7,0

### ✔️ Expected result
Shuffle+ to activate and give me that glorious shuffled action

### ❌ Actual result
Once Clicked nothing happens then the error in the image pops up and everything continues as usual

## 📷 Screenshots
image

iurimateus:
This PR contains two commits.

The first fixes #1152, where the request fails due to a missing field (with Queue returning undefined).
The second fixes an exception when the user has a empty folder, which breaks the recursive search.

### DIFF
--- a/Extensions/shuffle+.js
+++ b/Extensions/shuffle+.js
@@ -321,8 +321,9 @@
                 },
             });
         }
+
         await Spicetify.CosmosAsync.put("sp://player/v2/main/queue", {
-            queue_revision: Spicetify.Queue?.revision,
+            queue_revision: Spicetify.Queue?.queueRevision,
             next_tracks: list.map((uri) => ({
                 uri,
                 provider: context ? "context" : "queue",
@@ -330,7 +331,7 @@
                     is_queued: isQueue,
                 },
             })),
-            prev_tracks: Spicetify.Queue?.prev_tracks,
+            prev_tracks: Spicetify.Queue?.prevTracks,
         });
         success(count);
         Spicetify.Player.next();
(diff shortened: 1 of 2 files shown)


iurimateus:
> You are not handling the case when the result is returned empty. Or am I missing something?
I wasn't clear, `Queue.revision` (and the others) returns undefined because the field
doesn't exist, not because the internal function is returning that (or empty).

(No idea if/how we should handle missing fields).

> How does changing the names in global.d.ts file fixes the issue?

I'm guessing there was a change to the internal function, so 
`get Queue() {return Spicetify.Player.origin?._queue?._state}` now returns pascal case fields.

Running `Spicetify.Queue` on console returns 
```json
{
  queueRevision: "",
  nextTracks: [],
  prevTracks: [],
  track: {...},
}
```

Sure, the change in the interface doesn't do anything, but I'm also changing
the fields on the request.

Reproducing:

```js
let list = ['3vrPkeXFkClImJm5ZD3I4J']
Spicetify.CosmosAsync.put("sp://player/v2/main/queue", {
    queue_revision: Spicetify.Queue?.revision,
    next_tracks: list.map((uri) => ({
        uri,
        provider: "queue",
        metadata: {
            is_queued: false,
        },
    })),
    prev_tracks: Spicetify.Queue?.prevTracks,
});
```
yields 
```json
{
  "status": 400,
  "headers": {
    "error-description": "Missing 'queue_revision'."
  }
}
```

PR #1219, opened 2021-11-03
NEW: export spicetify version to js

### LINKED ISSUE #1202: Expose spicetify-cli version in Spicetify object
It would be nice to have the current spicetify-cli version exposed somewhere in the Spicetify object. For example in a theme I'm working on there is an option to copy the OS, Spotify and Theme versions to the clipboard to paste them in issues. The ability to have the current spicetify-cli version too would be nice as the user won't have to manually add it.

This could also be useful for themes to notify the user that the current spicetify-cli version is outdated.

---

If you'd like to see this feature implemented, add a 👍 reaction to this post.

JulienMaille:
FIX #1202

### DIFF
--- a/src/apply/apply.go
+++ b/src/apply/apply.go
@@ -86,6 +86,8 @@ func htmlMod(htmlPath string, flags Flag) {
 		helperHTML += `<script defer src="helper/homeConfig.js"></script>` + "\n"
 	}
 
+	helperHTML += `<script defer src="helper/version.js"></script>` + "\n"
+
 	for _, v := range flags.Extension {
 		if strings.HasSuffix(v, ".mjs") {
 			extensionsHTML += `<script defer type="module" src="extensions/` + v + `"></script>` + "\n"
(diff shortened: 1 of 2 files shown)


itsmeow:
Suggestion:
```js
helperHTML += `<script>Spicetify.version="${backupSpicetifyVersion}";</script>` + "\n"
```

OR
append said statement to `Apps/xpui/helper/spicetifyWrapper.js`, which is already loaded here
https://github.com/khanhas/spicetify-cli/blob/3375f50faebf0afe9f76d434b6418566478ec282/src/preprocess/preprocess.go#L151

JulienMaille:
@itsmeow better?

JulienMaille:
@afonsojramos what do you think?

afonsojramos:
@JulienMaille I'll look at it over the weekend, as I'll have to explore that code area a bit more, but it looks good!

itsmeow:
@JulienMaille See my review - you need to define SpicetifyVer as a string in the struct

itsmeow (approved):

PR #1222, opened 2021-11-04
Podcast Songs Extension

xXChampionsXx:
Just 1 simple line that starts every Song/Podcast from 0 (Basically only applies to Podcasts).
Thought maybe some other People also have Podcasts with only Songs in them...

### DIFF
--- a/Extensions/podcastSongs.js
+++ b/Extensions/podcastSongs.js
@@ -1,0 +1,13 @@
+// @ts-check
+// NAME: Podcast Songs
+// AUTHOR: xXChampionsXx
+// VERSION: 0.1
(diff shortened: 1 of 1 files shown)


CharlieS1103:
@xXChampionsXx Especially with an extension like this, it would be better to just publish it on your own, most extensions on the actuall GitHub are either more complex/useful or are useful for new devs to read through so they can make their own. If you'd like you can make it more discoverable with this though, https://github.com/CharlieS1103/spicetify-marketplace (it's in alpha phase however.)

xXChampionsXx:
Thanks for the suggestion I will look into it.

PR #1269, opened 2021-11-14
Open File explorer with -c and -config commands + Bug Fix

CharlieS1103:
With mac, the config file was changed to a hidden folder for cleanness purposes, however, users are typically unable to access those without following special steps (Which many inexperienced users may not be capable of.) This provides them with support. Also, before this commit -c and -config was printing the incorrect file path, this is now resolved.


CharlieS1103:
Now realize that for whatever reason my local files are not synced with Spicetify, will redo.

PR #1270, opened 2021-11-14
Open File explorer with -c and -config commands

CharlieS1103:
With mac, the config file was changed to a hidden folder for cleanness purposes, however, users are typically unable to access those without following special steps (Which many inexperienced users may not be capable of.) This provides them with support


itsmeow:
This is already implemented. See `spicetify config-dir`

CharlieS1103:
Aw

PR #1276, opened 2021-11-16
Rename update command and add update as alias for upgrade

CharlieS1103:


khanhas:
Nah. They're fine as they are.

PR #1301, opened 2021-11-25
Fix fullAppDisplay

borear:
Lyrics:
-fix active line blur,
-fix the artifact at the bottom of the page;

Background image: 
-fix blur


khanhas:
Merged active line blur fix.

PR #1325, opened 2021-11-30
Improve visibility and design of some elements in SpicetifyDefault theme

sagars007:
Looks like some elements have the wrong colors/not visible as some of the css selectors are different in the latest spotify version. I've fixed all the issues encountered during my testing. Looks good in all color schemes of SpicetifyDefault theme.

https://imgur.com/a/SYKMbN4


JulienMaille:
Thank for your PR. However, be aware that all rules using the random class names like `_XjIr3Jebi9c96f7h74J` will break on upcoming Spotify updates.
You want to find what their original class name should be and update https://github.com/khanhas/spicetify-cli/blob/master/css-map.json

sagars007:
Yeah you're right, this is only a temporary fix. It's annoying that the random class names keep changing every update, breaks many user.css stylesheets. How do we find the corresponding original class names? The random class names in my code are not available in the css map.

TheGeeKing:
They need to be mapped or you use a class above it.

sagars007:
> They need to be mapped or you use a class above it.

Yeah they are not mapped yet, but how do we find them to be mapped?

TheGeeKing:
We need to name them to something like others are named(same format), if you have any idea let me know and I can map them.

sagars007:
Oh ok that makes sense. I'll try to find them when I'm free later on. Seems like a tedious task

afonsojramos:
@sagars007 it sure is... 😢 But, at the moment, it is the only way unfortunately.

afonsojramos:
@sagars007 Can you update the PR with proper classnames? If not we will close this PR as it doesn't make sense to push it to master.

sagars007:
> @sagars007 Can you update the PR with proper classnames? If not we will close this PR as it doesn't make sense to push it to master.

Sure. I'll update the css map with the ones I can find

Issue #1355, opened 2021-12-10
keyboardShortcut.js addon for Linux users.


stankovictab:
Song skip and volume change shortcuts (CTRL + Arrow Keys) for the Linux version of Spotify for some reason doesn't work,\
and this edit can get around that issue.\
Sorry for not making a pull request I'm still new to GitHub 😅

```js
registerBind("ARROW_RIGHT", true, false, false, nextSong)
function nextSong() {
	document.querySelector(".main-skipForwardButton-button").click();
}
registerBind("ARROW_LEFT", true, false, false, previousSong)
function previousSong() {
	document.querySelector(".main-skipBackButton-button").click();
}
registerBind("ARROW_UP", true, false, false, increaseVolume)
function increaseVolume() {
	Spicetify.Player.origin.setVolume(Spicetify.Player.getVolume() + 0.1);
}
registerBind("ARROW_DOWN", true, false, false, decreaseVolume)
function decreaseVolume() {
	Spicetify.Player.origin.setVolume(Spicetify.Player.getVolume() - 0.1);
}
```

InteractiveNinja:
@stankovictab I added a pullrequest for your code, hopefully I didnt disregarded any rules ^^

stankovictab:
@InteractiveNinja Looks great, thanks!

PR #1377, opened 2021-12-18
Add shortcuts for nextSong previousSong increaseVolume decreaseVolume…

InteractiveNinja:
Code by @stankovictab Issue #1355

### DIFF
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -84,6 +84,16 @@
     // Forward Slash to open search page
     registerBind("/", false, false, false, openSearchPage);
 
+    // CTRL + Arrow Left Next and CTRL + Arrow Right  Previous Song
+
+    registerBind("ARROW_RIGHT", true, false, false, nextSong)
+    registerBind("ARROW_LEFT", true, false, false, previousSong)
+
+    // CTRL + Arrow Up Increase Volume CTRL + Arrow Down Decrease Volume
+
+    registerBind("ARROW_UP", true, false, false, increaseVolume)
+    registerBind("ARROW_DOWN", true, false, false, decreaseVolume)
+
     // F to activate Link Follow function
     const vim = new VimBind();
     registerBind("F", false, false, false, vim.activate.bind(vim));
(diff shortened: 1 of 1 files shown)


khanhas:
Do these default keyboard shortcuts actually not work on all Linux distros or just yours? Can I get some more input?
Also wrap them in a if condition checking OS, I'm sure on Windows and MacOS they work just fine.

stankovictab:
@khanhas That's right, they do work fine on Windows, \
however I know from experience and from my friends that the shortcut doesn't work on Ubuntu, Kubuntu, Linux Mint or Manjaro. \
It's been an active issue for a while now, you can see the issue [here on the Spotify Community forum](https://community.spotify.com/t5/Desktop-Linux/Forward-and-back-keyboard-shortcuts-no-longer-working/td-p/5257504). \
It appears to be the same [on Mac](https://community.spotify.com/t5/Desktop-Mac/Forward-and-back-keyboard-shortcut-feature-no-longer-working/m-p/5245215), but I haven't experimented with it.\
I'll add the OS check, it should work fine, but please correct it if it's wrong. 🙂

stankovictab:
This should do the trick.
```js
function nextSong() {
	if (window.navigator.userAgent.indexOf("Win") == -1) // Works only on Linux and Mac
		document.querySelector(".main-skipForwardButton-button").click();
	}

function previousSong() {
	if (window.navigator.userAgent.indexOf("Win") == -1) // Works only on Linux and Mac
		document.querySelector(".main-skipBackButton-button").click();
	}

function increaseVolume() {
	if (window.navigator.userAgent.indexOf("Win") == -1) // Works only on Linux and Mac
		Spicetify.Player.origin.setVolume(Spicetify.Player.getVolume() + 0.1);
	}

function decreaseVolume() {
	if (window.navigator.userAgent.indexOf("Win") == -1) // Works only on Linux and Mac
		Spicetify.Player.origin.setVolume(Spicetify.Player.getVolume() - 0.1);
	}
```

khanhas:
Okay then.
But please wrap functions and `Spicetify.KeyboardShortcut` calls in one if condition, not each of these like stankovictab's code.
We don't need to add these shortcuts handler for Windows since they work out of the box.

InteractiveNinja:
Do we really need to also check the function definitions?

PR #1389, opened 2021-12-31
Select parent element, so dribbblish's custom search box will be selected

liranco:
Dribbblish has its own search box, it would be nice if keyboardShortcut.js would select it instead.

### DIFF
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -165,7 +165,7 @@
      * @param {KeyboardEvent} event
      */
     function openSearchPage(event) {
-        const searchInput = document.querySelector(".main-topBar-topbarContentWrapper input");
+        const searchInput = document.querySelector(".main-topBar-container input");
         if (searchInput) {
             searchInput.focus();
         } else {


afonsojramos:
@liranco I'm unable to test this locally as I've never used `keyboardShortcut.js`. Can you give me some guidelines on the keymaps?

liranco:
Sure!

The script makes a bunch of calls to `registerBind`, each one contains a key binding, some flags, and a callback to be called whenever the key is pressed.

Currently, there's a default binding for the `/` key, which is bound to `openSearchPage`. 

On its original implementation, `openSearchPage` focused user input on Spotify's default search bar. However, dribbblish implements its own search box, and sets Spotify's one to hidden.

My modification changed a bit the css selector used to find the search box, so that dribbblish's search box will return first, if it exists.

If it doesn't exist (some other theme is applied), then the selector returns the original element, and so no change occurs.

Let me know if I can help with testing, further explanation, or anything else!

afonsojramos (approved):
Thanks for waiting and sorry for the time to approve, it went off my radar.

Will merge! 🚀 💯

PR #1393, opened 2022-01-05
Patch React element regex

CharlieS1103:


CharlieS1103:
ty to @itsmeow for showing me how

PR #1410, opened 2022-01-16
CSS map: update 'more albums by artist' shelf

GeorgeHahn:
From Spotify for Windows 1.1.76.447.g11f432d8


afonsojramos:
@itsmeow this has happened before right? Where we had two different keys for the same value?

@GeorgeHahn can you send a screenshot of the html hierarchy where this class is present?

GeorgeHahn:
@afonsojramos 
image

image

itsmeow:
Two keys is fine - khanhas wanted the old ones around so it's probably unused.

itsmeow (approved):

afonsojramos:
Thanks @itsmeow @GeorgeHahn!

PR #1416, opened 2022-01-21
Add `spicetify` to `$PATH` on linux

afonsojramos:
After opening https://github.com/afonsojramos/spicetify-docs/pull/11 the question of whether or not adding to path on Linux was happening at the `install.sh` script was raised. After a quick look at it, it did seem that it was not happening.

This PR, in theory, fixes it, however, I do not have a full linux machine so if anyone could test this I'd be grateful, otherwise, I'll try to test on a VM, but I haven't had much time lately.

I took two Stack Overflow threads into consideration ([#1](https://stackoverflow.com/a/20054809) and [#2](https://superuser.com/a/183980)), which is why I set the path in `.profile`.

Tagging @theRealPadster and @CharlieS1103 for visibility.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,4 +1,4 @@
-# Copyright 2019 khanhas. GPL license.
+# Copyright 2022 khanhas. GPL license.
 # Edited from project Denoland install script (https://github.com/denoland/deno_install)
 param (
   [string] $version
--- a/install.sh
+++ b/install.sh
@@ -31,8 +31,8 @@ fi
 spicetify_install="$HOME/.spicetify"
 path="export PATH=\"\$PATH:\$HOME/.spicetify\""
 
-if [[ -f "$HOME/.bashrc" ]] && ! grep -q "$path" "$HOME/.bashrc"; then echo "${path}" >>"$HOME/.bashrc"; fi
-if [[ -f "$HOME/.zshrc" ]] && ! grep -q "$path" "$HOME/.zshrc"; then echo "${path}" >>"$HOME/.zshrc"; fi
+if [[ -f "$HOME/.bashrc" ]] && ! grep -q "$path" "$HOME/.bashrc"; then echo "${path}" >>"$HOME/.bashrc" && echo "SAVING         ${spicetify_install} to \$PATH"; fi
+if [[ -f "$HOME/.zshrc" ]] && ! grep -q "$path" "$HOME/.zshrc"; then echo "${path}" >>"$HOME/.zshrc" && echo "SAVING         ${spicetify_install} to \$PATH"; fi
 
 exe="$spicetify_install/spicetify"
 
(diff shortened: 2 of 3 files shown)


theRealPadster:
> On modern unices, there's an added complication related to ~/.profile. If you log in in a graphical environment (that is, if the program where you type your password is running in graphics mode), you don't automatically get a login shell that reads ~/.profile. Depending on the graphical login program, on the window manager or desktop environment you run afterwards, and on how your distribution configured these programs, your ~/.profile may or may not be read. If it's not, there's usually another place where you can define environment variables and programs to launch when you log in, but there is unfortunately no standard location.

The notes about login shells are interesting. The terminal app I use, Tilix, has a warning screen that pops up when you're not running it as a login shell, but I don't think most do this. 

I'll do a little testing, with the default GNOME terminal etc a bit later today on my PC and let you know what I find.

afonsojramos:
I knew about the different `.${shellname}rc`'s that are around, but did not realise it was such a hard issue... I suppose that this is another occurrence where the lack of "standards" damages the end product, in this case Linux.

theRealPadster:
Yeah it's not great. I have a shell env file that I `source` in multiple shell-rc files so I have the same environment when playing with different shells. I think it's usually safe to just use .bashrc because most people have or use bash, and it's the most common. Might be worth checking if it's possible to see what shell the script is being executed by and changing the RC file based on that

theRealPadster:
@afonsojramos  I don't have a `.profile`, so I added one with "echo 'profile'" inside. It never runs, even in Tilix with the login shell enabled. Added "echo 'bashrc'" to my `.bashrc`, and it runs when I run bash, but not when I run zsh or Fish (my default shell). 

I think bash is the most common default on Linux (and some distros might even have their shell rc file set to source the bashrc), so that might be the best option? If I manually installed and set up another shell, like Fish, then I probably have enough basic understanding to add spicetify to my `$PATH` (which the spicetify install script mentions if it has issues, I believe). 

If you find a little time, testing in a VM might also be a good idea, or if anyone else is able to test or offer any input, that would be appreciated.

afonsojramos:
@theRealPadster thanks, great input! What I think I'll do, is what I was already considering before, which is to check which files exist and append to each one of them. So, even if the user has two shells, it can use it on both!

theRealPadster:
@afonsojramos Ah, good call. A bit more complicated, but probably the best solution. My other thought was to use something like `$0`, which would output the shell name, for POSIX-compliant shells like bash or zsh. This doesn't work for Fish though (which I realise nobody uses :P).

afonsojramos:
So, deprecated the use of `zshenv` since `Spicetify` is only a CLI, so adding it in `.zshrc` should suffice. Reference [here](https://unix.stackexchange.com/questions/71253/what-should-shouldnt-go-in-zshenv-zshrc-zlogin-zprofile-zlogout).

PR #1419, opened 2022-01-21
Fixes for expFeatures.js

jeroentvb:
This PR fixes an error caused by setting a property on `activeProperties` that doesn't exist (anymore?). The property was retrieved from localStorage. In my case the key was '58'.

It also fixes https://github.com/khanhas/spicetify-cli/issues/1417. The experimental features didn't render anymore, because of outdated use of properties on the `resolver` object.

I was testing with the 'enhance playlist' feature, and noticed toggling the feature required a reload to show up in the UI. Would be nice to automatically reload Spotify on closing the popup window, but I don't see a way to detect the window closing.
Also the 'Show "Made For You" entry point in the left sidebar' feature seems to be broken, not sure if we can do anything about that.

### DIFF
--- a/jsHelper/expFeatures.js
+++ b/jsHelper/expFeatures.js
@@ -106,12 +106,12 @@ button.switch[disabled] {
             return container;
         }
 
-        for (const propName in resolver.propertySet.properties) {
-            const prop = resolver.propertySet.properties[propName].spec;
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):
Awesome! Thanks for the PR!

Issue #1406, opened 2022-01-13
Keyboardshortcuts "ctrl+q" error


XanoPhobious:
## ℹ Computer information

- Spotify version: 1.1.76.447.g11f432d8
- Spicetify version: 2.8.4

## 📝 Provide detailed reproduction steps (if any)

1. Run Spotify
2. Press Ctrl+Q
3. The shortcut does not display "Queue"

### ✔️ Expected result

The queue list will be displayed

### ❌ Actual result

The keyboard shortcut (ctrl+q) does nothing.

## 📷 Screenshots

![Clip](https://user-images.githubusercontent.com/97661288/149287406-9c193fd5-77fd-4079-824c-a07f8219ae30.jpg)

XanoPhobious:
This issue is back. I don't know if the cause the same.

PR #1423, opened 2022-01-22
fix keyboard shortcut for queue button

ajiou:
fix issue #1406

### DIFF
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -109,7 +109,7 @@
     }
 
     function clickQueueButton() {
-        document.querySelector(".control-button-wrapper .spoticon-queue-16").click();
+        document.querySelector(".control-button-wrapper > button.Button-sc-1dqy6lx-0").click();
     }
 
     function clickNavigatingBackButton() {
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -109,7 +109,7 @@
     }
 
     function clickQueueButton() {
-        document.querySelector(".main-nowPlayingBar-right > div.ExtraControls > div.GlueDropTarget > div.control-button-wrapper > button").click();
+        document.querySelector("div.ExtraControls div.control-button-wrapper > button").click();
     }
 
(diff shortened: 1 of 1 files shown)


Extensions/keyboardShortcut.js
```diff
@@ -109,7 +109,7 @@
     }
 
     function clickQueueButton() {
-        document.querySelector(".control-button-wrapper .spoticon-queue-16").click();
+        document.querySelector(".control-button-wrapper > button.Button-sc-1dqy6lx-0").click();
```
afonsojramos, line 112: It seems that the class being used is not mapped, so let's try to use a generic way of accessing this button, ie, simply this:
```js
document.querySelector(".control-button-wrapper > button").click();
```

afonsojramos (changes requested):
@ajiou I've never used keyboard shortcuts, can you tell me how to test this?

Extensions/keyboardShortcut.js
ajiou: I see. Also, I discovered that this selector would select the "Remove" button first, so i make it a bit more specific.

afonsojramos: The remove button? Which remove button? If that's the case then maybe we do need to be more specific.

ajiou: The one on the `.main-nowPlayingBar-left` to make the auto-suggested song not to play again
image

Extensions/keyboardShortcut.js
```diff
@@ -109,7 +109,7 @@
     }
 
     function clickQueueButton() {
-        document.querySelector(".control-button-wrapper .spoticon-queue-16").click();
+        document.querySelector(".main-nowPlayingBar-right > div.ExtraControls > div.GlueDropTarget > div.control-button-wrapper > button").click();
```
afonsojramos, line 112: Let's keep it simple to reduce our dependency on the CSS Mappings: `div.ExtraControls div.control-button-wrapper > button`

afonsojramos (changes requested):

afonsojramos (approved):

PR #1424, opened 2022-01-23
Spicetify.Player volume method fixes/replacements

machinemessiah:
Switched out the missing/removed Spicetify.Player volume methods (Spicetify.Player.origin._volume no longer exists) for the new methods exposed by Spicetify.Platform.PlaybackAPI. The volume events previously dispatched by Spicetify.Player.origin._events are now dispatched by Spicetify.Platform.PlaybackAPI._events, which means extensions which listen to volume events will need to be modified (or someone else can come up with a much better way to sort this volume bug out like creating an alias or something...).

VScode and diff checkers online both confirm that the only changes are the four methods I replaced (lines 33-36), but git insists I changed every single line...I'm an idiot so I don't know how to fix that...

I've tested, but I have no idea if this is the correct route or whether this is just a band-aid and larger changes ultimately need to be made to Spicetify.Player.


afonsojramos:
Can you check your Line Endings? It is very likely that you saved it as CRLF by accident 🤔

machinemessiah:
> Diff tool will not show proper changes until line endings are fixed - cannot review this atm

I _think_ I've corrected the formatting.

How would you feel about adding a .gitattributes file so this sort of issue never happens - regardless of the contributor's OS (AFAIK this happened because I'm on windows which defaults to CRLF). 

I barely know what I'm doing when it comes to git, [but this solution came up after a quick google search](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#a-simple-gitattributes-config).

nodgear:
Line endings are still poo poo

also, when installing git on windows the default option is to write with windows ending and commit with unix ones.

CharlieS1103:
This seems all good to me, i'm not very knowledgable on OS's and line ending and things like that however.

afonsojramos (approved):
Merging this as it works and is required to fix a lot of WNP issues.

PR #1425, opened 2022-01-23
CSS map: update `main-rootlist-wrapper`

SendNukez:
This class is broken since spotify `v1.1.76.447`, and the cause of https://github.com/JulienMaille/dribbblish-dynamic-theme/issues/153

Here is a screenshot of the DOM tree:
image

I'd also map the child elements, but I don't know the actual class names since they're not used in my scripts.


afonsojramos (approved):

afonsojramos:
Thanks @SendNukez!

PR #1426, opened 2022-01-24
Add `spicetify` to `$PATH` for Fish shell

theRealPadster:
This adds `spicetify` to the `$PATH` for the Fish shell. 
I also added the shell to the echo line, for people who have more than one shell rc file. This way it doesn't output the same message multiple times, which could be confusing. 

Reference: https://fishshell.com/docs/current/faq.html#how-do-i-run-a-command-every-login-what-s-fish-s-equivalent-to-bashrc-or-profile

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -31,9 +31,9 @@ fi
 spicetify_install="$HOME/.spicetify"
 path="export PATH=\"\$PATH:\$HOME/.spicetify\""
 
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):

afonsojramos:
So it is possible afterall! Thanks @theRealPadster!

Issue #1422, opened 2022-01-22
lyrics.providedBy text bug on Lyrics plus. (Latest Spotify update) 


IamNickerson:
## ℹ Computer information

- Spotify version: 1.1.77.643
- Spicetify version: 2.8.4

## 📝 Provide detailed reproduction steps (if any)

1.Open Spotify
2..Try to update Spotify to version 1.1.77.643
3.Have the custom app lyrics plus installed in spicetify
4.Do a spicetify apply
5.Play a song
6.Go the tab Lyrics.
This the expected result will be:
![image](https://user-images.githubusercontent.com/61643704/150628096-acd054ee-0a4c-410c-9a69-fe8b8be44950.png)

PR #1433, opened 2022-01-28
fix: lyrics.providedBy text bug on lyrics-plus

huhridge:
fixed lyrics.providedBy text bug on Lyrics plus.  #1422


afonsojramos (approved):
Thanks for the fix! 🎉

PR #1439, opened 2022-01-30
Automatically append .js (if absent) to extensions in config

CharlieS1103:
Should make extension installation process slightly simpler for new users


src/cmd/apply.go
```diff
@@ -245,7 +245,10 @@ func pushExtensions(destExt string, list ...string) {
 			extPath = v
 		} else {
 			extName = v
-			extPath, err = utils.GetExtensionPath(v)
+			if(!strings.Contains(extName,".")){
```
afonsojramos, line 248: I'd say to verify if it has `".js"` instead since a file name can have a . without it being the extension.

afonsojramos (changes requested):

src/cmd/apply.go
CharlieS1103: @afonsojramos Done.

afonsojramos (approved):

Issue #1443, opened 2022-01-31
LyricsPlus Genius - Twin Lyrics-Page Issue


DanBeast1337:
## ℹ Computer information

- Spotify version: 1.1.77.643.g3c4c6fc6
- Spicetify version: 2.8.5

## 📝 Provide detailed reproduction steps (if any)

When opening Lyrics Plus and going to the Genius page, it shows the lyrics twice. 
Depending on the size, you can see more or less of the right part. 
It can show one lyric-page on the left, and a translated one on the right side. 

Not sure if this is wanted, but it definitely does not seem like it, or properly formatted. 

<sub>(don't mind the lyrics below, was to lazy to change it to english lmao)</sub>

## 📷 Screenshots

_Are there any useful screenshots?_

<sub>**vv**</sub> smaller window, windowed (1080p, 100% scaling) <sub>**vv**</sub>
![image](https://user-images.githubusercontent.com/9980963/151804838-3732aefb-e704-4fd8-936a-3734a21864a7.png)

<sub>**vv**</sub> maximized window (1080p, 100% scaling)(bottom right is just the previous screenshot, thanks sharex) <sub>**vv**</sub>
![image](https://user-images.githubusercontent.com/9980963/151804831-395bbdae-f4b1-4130-baa5-736c5a53fb0f.png)

afonsojramos:
Will close as this has been fixed in #1444.

PR #1444, opened 2022-01-31
temporarily disable split display for genius

ajiou:
temporary fix for #1443


ajiou:
Don't know why the author did this, so I just made the statement always false to keep the original code.

CharlieS1103:
Ty this has been annoying me for a while

afonsojramos (approved):

PR #1445, opened 2022-02-01
fix: replaced file existence check for shell check

### LINKED ISSUE #1435: macOS installer is broken
## ℹ Computer information
Monterey 12.3 Beta 

- Spotify version: 1.1.77
- Spicetify version: 2.8.4

Restarting terminal doesnt work anymore 

I even tried manually downloading install.sh unfortunately no luck for me 

Bildschirmfoto 2022-01-29 um 12 00 25

afonsojramos:
In theory, fixes #1435.

@ririxidev @theRealPadster

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -31,9 +31,9 @@ fi
 spicetify_install="$HOME/.spicetify"
 path="export PATH=\"\$PATH:\$HOME/.spicetify\""
 
-if [[ -f "$HOME/.bashrc" ]] && ! grep -q "$path" "$HOME/.bashrc"; then echo "${path}" >>"$HOME/.bashrc" && echo "SAVING         ${spicetify_install} to \$PATH (bash)"; fi
-if [[ -f "$HOME/.zshrc" ]] && ! grep -q "$path" "$HOME/.zshrc"; then echo "${path}" >>"$HOME/.zshrc" && echo "SAVING         ${spicetify_install} to \$PATH (zsh)"; fi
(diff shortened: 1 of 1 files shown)


theRealPadster:
`$SHELL` doesn't work in Fish. For that matter, the rest of the syntax isn't compatible with Fish anyways, so the script can not be run with Fish. (You'd need to run under a different shell, but it would be nice if `spicetify` was available in the `$PATH` afterwards.) I believe that `$SHELL` is meant to output the currently running shell, which would mean that it only adds to the `$PATH` of a single shell, which we don't want.

afonsojramos:
> the rest of the syntax isn't compatible with Fish anyways

I thought #1426 made it work, what do you mean? Or are you simply referring to `if [[ -f "$HOME/.config/fish/config.fish" ]]`?  I can roll that back.

> which would mean that it only adds to the $PATH of a single shell, which we don't want.

I disagree here. This should be the default behaviour.

theRealPadster:
That commit enables it to add to the fish path, but the script itself will not run under fish because fish isn't posix-compatible and the syntax isn't identical. 
I don't see the harm in adding it to multiple shells if the user has multiple shells set up. I generally use fish, but have zsh on MacOS as well, and have experimented with some different ones. This is why I think it's still a net positive to include support for the fish path even though the script can't be run from fish

afonsojramos:
In that case I'll rollback the fish change. Either way, the installation must be run in a POSIX compliant shell, so it will then work with fish after installation.

afonsojramos:
@theRealPadster now we should be good to cover all edge cases.

theRealPadster:
> @theRealPadster now we should be good to cover all edge cases.

Awesome, looks good to me, thanks!

ririxidev:
Yeah looks good now ig

Issue #1369, opened 2021-12-12
(LyricsPlus) can show Instrumental tag


SupLiCatinG:
## 📝 Provide a description of the new feature

![image](https://user-images.githubusercontent.com/94017480/145731069-076caf6d-ae46-4755-9450-47f98f3f2a04.png)

popupLyrics can show that why Lyrics Plus can't show the Instrumental tag?

If you'd like to see this feature implemented, add a 👍 reaction to this post.

PR #1446, opened 2022-02-01
added Instrumental to lyrics-plus, and ability to play stations with auto-shuffle on fixing #1448

huhridge:
added #1369

Note: Even though I have added instrumental for both Netease and musixmatch, I would recommend putting musixmatch over netease in provider order.

also added the case where if a provider only has unsynced lyrics, rest are also gone through to check if there are synced ones.


afonsojramos:
Wow, thanks for the PR, I'll have to test it a bit and will get back to you soon!

huhridge:
Please do test it, I did as much as I could, and couldn't find anything wrong, but I may have missed something

huhridge:
Fixed: #1448
Since there is no simple way, to get the queue of a radio, added a workaround to play with original play function.

CustomApps/lyrics-plus/ProviderNetease.js
```diff
@@ -120,6 +121,9 @@ const ProviderNetease = (function () {
         const lyrics = lines
             .map((line) => {
                 const { time, text } = parseTimestamp(line);
+                if (text === "纯音乐, 请欣赏") {
+                    isInstrumental = true;
+                }
```
afonsojramos, line 126: What is this exactly? 😅

afonsojramos (approved):
Thanks for all the fixes!! Looking good!

Extensions/shuffle+.js
```diff
@@ -116,6 +116,10 @@
             case Spicetify.URI.Type.TRACK:
             case Spicetify.URI.Type.EPISODE:
                 return [uri];
+            case Spicetify.URI.Type.STATION:
+            case Spicetify.URI.Type.RADIO:
+                playerPlayOGFunc({ uri: uri }, { featureVersion: Spicetify.Platform.PlayerAPI._defaultFeatureVersion });
+                return ["playedstation"];
```
afonsojramos, line 122: It is a workaround, but I think it is very valid.
Do we need this line though? 
```rust
playerPlayOGFunc({ uri: uri }, { featureVersion: Spicetify.Platform.PlayerAPI._defaultFeatureVersion });
```

CustomApps/lyrics-plus/ProviderNetease.js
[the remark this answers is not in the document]
huhridge: yeah, that is the only way to detect instrumental in netease, the Musixmatch one will probably get used more often

Extensions/shuffle+.js
huhridge: Yeah, this is the only way to play stations, I tested it out a lot of different ones, but this is the only one that works.

**For a bit of context**: 
As I am sure you're aware, shuffle+ injects its own playing function when auto-shuffle is on, thankfully it does store the original function, which is, as the name suggests `playerPlayOGFunc`, so I have to use this to play the stations. And those arguments is the only way I know how to call it successfully.

afonsojramos: I was actually not aware when I made the comment as I haven't messed around with shuffle+, but afterwards I went digging into the code and found that binding. Thanks for the context!!

PR #1457, opened 2022-02-05
fix: issue where new `Spicetify.ContextMenu` items were not added to the now playing song in the bottom.

huhridge:
For the now playing song, the properties of the event were a bit changed, so added an exception.


afonsojramos (approved):
I had no idea about this code segment and it looks totally wacky, but I love how the glue works 💯

PR #1495, opened 2022-02-21
Changed selector for focusOnApp() in keyboardShortcuts.js

Kafva:
I noticed that the keymaps for scrolling stopped working in the current release.
 
- Spotify version: Spotify for macOS (Apple Silicon) 1.1.79.763.gc2965cdf
- Spicetify version: 2.9.1

The current version of `keyboardShortcut.js` uses the selector `main .os-viewport` (which no longer exists) to focus the main container before scrolling, switching the selector to `.Root__main-view .os-viewport` solved the issue for me.
Using only `.os-viewport` as the selector can be to general since it also matches the sidebar.

Screenshot 2022-02-21 at 21 32 40

### DIFF
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -205,7 +205,7 @@
     }
 
     function focusOnApp() {
-        return document.querySelector("main .os-viewport");
+        return document.querySelector(".Root__main-view .os-viewport");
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):
Validated on my machine! Thanks!

PR #1529, opened 2022-03-07
fix: patch part of menu hook

CharlieS1103:
@afonsojramos Created a seperate branch for continuous patching of the hooks (overtime)


afonsojramos:
@CharlieS1103 any news on this? Also, it seems that the last Regex has broken retrocompatibility, is it possible to readd it in some way?

afonsojramos (approved):

PR #1533, opened 2022-03-16
add path for void linux

wael444:
adding check in path for `/usr/libexec/spotify`: [reason](https://github.com/void-linux/void-packages/blob/a10403d2ac194303421ddfbdbd33c770cd4afacb/srcpkgs/spotify/template#L44)

isn't https://github.com/spicetify/spicetify-cli/blob/e7ea852355d55eb153d3d8e7279e714a883d1c55/install.sh#L36 a bashism? 
[install.sh](https://github.com/spicetify/spicetify-cli/blob/master/install.sh) is specified to use `sh` but in some systems `sh` is symlinked to `dash`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # Copyright 2022 khanhas. GPL license.
 # Edited from project Denoland install script (https://github.com/denoland/deno_install)
 
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # Copyright 2022 khanhas. GPL license.
 # Edited from project Denoland install script (https://github.com/denoland/deno_install)
 
--- a/src/utils/config.go
+++ b/src/utils/config.go
@@ -293,6 +293,7 @@ func linuxApp() string {
 	potentialList := []string{
 		"/opt/spotify/",
 		"/usr/share/spotify/",
+		"/usr/libexec/spotify/",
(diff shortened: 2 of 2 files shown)


src/utils/config.go
```diff
@@ -293,6 +293,7 @@ func linuxApp() string {
 	potentialList := []string{
 		"/opt/spotify/",
 		"/usr/share/spotify/",
+		"/usr/libexec/spotify/",
```
afonsojramos, line 296: Thanks for the Void Linux support! Had never heard of that distro before.

install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
afonsojramos, line 1: I don't think this is needed as `bash` is a superset of `sh`. Additionally, I'm not sure that our script is supported by bash (as it has some different syntax in some situations).

afonsojramos (commented):
Regarding the "bashism", word that I had never heard before 😂, this was the way that we found to support the different shells that are around. If you find a better way, please do open another PR for it!!

install.sh
wael444: like i said, [[ is a bashism. it is infact posix compliant and backwards compatible with posix scripts but when a bash script with bash features (such as [[) is run within a posix shell, it will error.

afonsojramos: which is why we use `sh`, to ensure maximum compatibility.

wael444: then the script should be rewritten in POSIX in that case.

wael444: opening new PR for a script rewritten in POSIX.

PR #1535, opened 2022-03-17
restructure install.sh

wael444:
changes:
+ remove ability to automatically add spicetify to shell, it should be done manually due to some users having different PATHs for shells. though it can still be added back if needed.
+ add check for `tar` and `curl` for safety.
+ change spicetify installation patch from `$HOME/.spicetify` to `$HOME/.local/share/spotify`: [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).


install.sh
```diff
@@ -1,75 +1,57 @@
...
-if (grep -q "zsh" $SHELL || [[ -f "$HOME/.zshrc" ]]) && ! grep -q "$path" "$HOME/.zshrc"; then echo "${path}" >>"$HOME/.zshrc" && echo "SAVING         ${spicetify_install} to \$PATH (zsh)"; fi
-if [[ -f "$HOME/.config/fish/config.fish" ]] && ! grep -q "$path" "$HOME/.config/fish/config.fish"; then echo "${path}" >>"$HOME/.config/fish/config.fish" && echo "SAVING         ${spicetify_install} to \$PATH (fish)"; fi
+# download uri
+shortcut=https://github.com/spicetify/spicetify-cli
+tag=$(curl -LsH 'Accept: application/json' $shortcut/releases/latest)
+tag=${tag%\,\"update_url*}
+tag=${tag##*tag_name\":\"}
+tag=${tag%\"}
+download_uri="$shortcut/releases/download/$tag/spicetify-${tag#v}-$target.tar.gz"
+unset tag
```
afonsojramos, line 24: Aren't we adding extra complexity unnecessarily? We don't need to know the tag

afonsojramos (commented):
Other than what I already said, the PR looks good 👌

install.sh
wael444: is there another way to get the latest release? your script before this had a check for the tag as well.

wael444: https://github.com/spicetify/spicetify-cli/blob/761ab44ca4d5b4b06b4937f11897b29a6f3c30bc/install.sh#L22

afonsojramos: But it ignored the tag 🤷 it just went for the latest and that's it! But I guess your approach is actually more readable!

install.sh
```diff
@@ -1,75 +1,57 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
afonsojramos: Well, I guess it doesn't make much difference 🤷  But this is better than it was so 👌

afonsojramos (changes requested):
Okay looking good then, please just apply the changes that I detailed [here](https://github.com/spicetify/spicetify-cli/pull/1535#issuecomment-1077736579).

wael444:
> Okay looking good then, please just apply the changes that I detailed [here](https://github.com/spicetify/spicetify-cli/pull/1535#issuecomment-1077736579).

sorry, i don't quite understand.
the spicetify binary has the tag name in the file name. it would still require getting the tag, or am i wrong?
if so, could you provide me with an alternative way to get the latest release then?

> I think that here we should still add it automatically since it is not fair to assume that every Linux user knows what the PATH is. Specially with SteamOS around the block, there should be an influx of Linux users that will not know what everything is.

alright, adding it back i suppose. ~~with a check to see if it exists first!~~

> Now, here we'll have to disagree because .local/share is mostly user-data specific folder, instead of a folder to install x program to. I can find you a list of examples that do this with a simple ls -a at ~.

well, i just like to keep the home directory clean! but it would mean going against the XDG Specifications.
i mean, isnt 
```
drwxr-xr-x     - wael CustomApps
drwxr-xr-x     - wael Extensions
drwxr-xr-x     - wael Themes
drwxr-xr-x     - wael jsHelper
.rwxr-xr-x 103Ki wael css-map.json
.rwxr-xr-x  44Ki wael globals.d.ts
```
mostly spicetify data?
but alright. reverting to `~/.spicetify` for an installation.

wael444:
i have added the ability to add to shell profile but, i'm pretty sure i made it much worse
it will need some testing however.

ririxidev:
> well, i just like to keep the home directory clean! but it would mean going against the XDG Specifications.
i mean, isnt

Problem with this is, this script is used for installation on macOS as well. If you really want to follow XDG Base Directory Specification, you would need to make exception for macOS (~/Library/Application Support/spicetify-cli). XDG Base Directory Specification doesn't exist on macOS, so it's your own choice and your own intention if you want make script compatible with it or not.

wael444:
> this script is used for installation on macOS as well

oh. i'm sorry, i didn't know about that.

> If you really want to follow XDG Base Directory Specification

the primary maintainer disagreed with the idea. so i can't do much about it.
least i can do for myself is move `.spicetify` manually and add the changes to my shell profile.

>  so it's your own choice if you want make script compatible with it or not.

i don't primarily have a MacOS machine to test so any MacOS users could modify the script themselves and initiate a PR.

afonsojramos (approved):
Looks good 🚀  Just add linux-arm64 if you can and I'll merge!

install.sh
```diff
@@ -1,75 +1,80 @@
...
-"Darwin arm64") target="darwin-arm64" ;;
-"Linux x86_64") target="linux-amd64" ;;
-*)
-	echo "Unsupported platform $(uname -sm). Only Darwin x86_64, Darwin arm64 and Linux x86_64 binaries are available."
-	exit
-	;;
+	"Darwin x86_64") target="darwin-amd64" ;;
+	"Darwin arm64") target="darwin-arm64" ;;
+	"Linux x86_64") target="linux-amd64" ;;
+	*) echo "Unsupported platform $(uname -sm). Only Darwin x86_64, Darwin arm64 and Linux x86_64 binaries are available."; exit ;;
```
afonsojramos, line 11: We actually already support linux-arm64 in the upgrade command, so I'm not sure why we are ignoring it 🤷

wael444: because i hadn't seen it. 
also, i'm not sure what the exact `arm64` in uname is called but ill just make it `Linux arm64`.

install.sh
```diff
@@ -8,7 +8,8 @@ case $(uname -sm) in
 	"Darwin x86_64") target="darwin-amd64" ;;
 	"Darwin arm64") target="darwin-arm64" ;;
 	"Linux x86_64") target="linux-amd64" ;;
-	*) echo "Unsupported platform $(uname -sm). Only Darwin x86_64, Darwin arm64 and Linux x86_64 binaries are available."; exit ;;
+	"Linux arm64") target="linux-arm64" ;;
```
ririxidev, line 11: `Linux aarch64` since you didn't know how it's called

wael444: oh, thanks

afonsojramos (approved):

PR #1543, opened 2022-03-21
fix: `fullAppDisplay` album info & background

huhridge:
As the hermes protocol is broken `1.1.81` onwards, this can "fix" the issue. Now the lyrics-plus background will always match spotify lyrics background.

Maybe, in the future, there is a way to restore the old colorextractor.

### DIFF
--- a/CustomApps/lyrics-plus/index.js
+++ b/CustomApps/lyrics-plus/index.js
@@ -162,9 +162,6 @@ class LyricsContainer extends react.Component {
     async fetchColors(uri) {
         let prominent = 0;
         try {
-            // const colors = await CosmosAsync.get(`hm://colorextractor/v1/extract-presets?uri=${uri}&format=json`);
-            // prominent = colors.entries[0].color_swatches[4].color;
-            // hermes deprecated 1.1.81 onwards
             const colors = await CosmosAsync.get(
                 `https://spclient.wg.spotify.com/color-lyrics/v2/track/${uri.split(":")[2]}/?format=json&vocalRemoval=false&market=from_token`
             );
--- a/CustomApps/lyrics-plus/index.js
+++ b/CustomApps/lyrics-plus/index.js
@@ -162,8 +162,13 @@ class LyricsContainer extends react.Component {
     async fetchColors(uri) {
         let prominent = 0;
         try {
-            const colors = await CosmosAsync.get(`hm://colorextractor/v1/extract-presets?uri=${uri}&format=json`);
-            prominent = colors.entries[0].color_swatches[4].color;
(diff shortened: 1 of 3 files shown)


huhridge:
NOTE: It only works for songs which also have lyrics on spotify, but imo it is better than nothing, i will find a better solution when i have some time

CustomApps/lyrics-plus/index.js
```diff
@@ -162,8 +162,13 @@ class LyricsContainer extends react.Component {
     async fetchColors(uri) {
         let prominent = 0;
         try {
-            const colors = await CosmosAsync.get(`hm://colorextractor/v1/extract-presets?uri=${uri}&format=json`);
-            prominent = colors.entries[0].color_swatches[4].color;
+            // const colors = await CosmosAsync.get(`hm://colorextractor/v1/extract-presets?uri=${uri}&format=json`);
+            // prominent = colors.entries[0].color_swatches[4].color;
+            // hermes deprecated 1.1.81 onwards
```
afonsojramos, line 167: Why did you choose to leave this commented and not delete it?

Extensions/fullAppDisplay.js
```diff
@@ -341,9 +341,13 @@ body.video-full-screen.video-full-screen--hide-ui {
 
         async getAlbumDate(uri) {
             const id = uri.replace("spotify:album:", "");
-            const albumInfo = await Spicetify.CosmosAsync.get(`hm://album/v1/album-app/album/${id}/desktop`);
+            // const albumInfo = await Spicetify.CosmosAsync.get(`hm://album/v1/album-app/album/${id}/desktop`);
+            // const albumDate = new Date(albumInfo.year, (albumInfo.month || 1) - 1, albumInfo.day || 0);
+            // hermes protocol deprecated 1.1.81 onwards
```
afonsojramos, line 346: Same here.

afonsojramos (approved):
Overall I agree with the implementations that you propose. I would even say that it is not a temporary solution, since it will probably work more reliably than the previous one!

If we need the previous code, we'll check the git history, no need to leave old code commented 😉

CustomApps/lyrics-plus/index.js
huhridge: That was just me being lazy, so I wouldn't forget how it worked before, should I remove it?

afonsojramos: Yeah, as I said on the review: "If we need the previous code, we'll check the git history, no need to leave old code commented 😉"

afonsojramos (approved):

PR #1548, opened 2022-03-25
Menu hook and Menu Item Hook patches

CharlieS1103:


itsmeow:
The first hook is not backwards compatible, but I honestly would not be concerned about it too much.

itsmeow (approved):

CharlieS1103:
@afonsojramos All good to merge in?

afonsojramos:
Thanks! Just had access to a PC now 😅 Just released a new version with this changes!

PR #1560, opened 2022-03-29
fix(fullAppDisplay): background for full screen display on a vertical monitor

kwerdenker:
**Issue:**
If the canvas size of the fullAppDisplay.js is higher then wide the background offset calculation goes wrong and the background is displayed with a gap towards the top.

**Fix**:
This PR changes the calculation to use a different offset in the case of `height > width` which corrects the issue and leads to an expected result.

**Comparison:**
<p float="middle">
spicetify_fullAppDisplay_vertical&nbsp;spicetify_fullAppDisplay_vertical_fixed
</p>


afonsojramos (approved):
Thank you for this PR! I have had this problem for a while, but never truly tried to fix it.

PR #1571, opened 2022-04-06
fix(windows): add missing prefs message

CharlieS1103:
Mostly just adding some error handling which should slim down the number of cases we see of this reported majorly.


src/utils/config.go
```diff
@@ -187,8 +187,10 @@ func FindPrefFilePath() string {
 	switch runtime.GOOS {
 	case "windows":
 		path := winPrefs()
-		if len(path) == 0 {
+		if len(path) == 0 && winXApp() != 0{
 			path = winXPrefs()
+		}else if(len(path) == 0){
+			PrintError("No valid path options found, ensure you have Spotify installed and have ran it for at least 30 seconds.")
```
afonsojramos, line 193: First of all, there is a type mismatch here as `winXApp()` returns a `string`. Second of all, we already [print an error in case the path is not set](https://github.com/spicetify/spicetify-cli/blob/2b4f2e1df65b5d0a5d0e9f77cda5f1120be12b6f/src/utils/config.go#L136), so why are we adding another check for the same thing?

afonsojramos (changes requested):

src/utils/config.go
CharlieS1103: To be entirely honest, didn't really check this code just assumed it would work after typing it out in github editor online(can't really test it regardless since im not on windows) but pretty much the prefs file only generates if the user has opened spotify at least once, and if the user sets the prefs path in spicetify AND THEN opens spotify for the first time. It will default to the windows store pref path. This reverses that process.

CharlieS1103: @afonsojramos Actually, now that i've had time to review, in golang you are able to use 0 to represent a string that has no contents, so the type mismatch isn't really a type mismatch, and to further explain myself, pretty much, we have a lot of people who are coming off of new spotify installs, and don't yet have a prefs path configured. Unfortunately, in the previous system, that would result in their prefs path being hardcoded to the windows store version EVEN IF they did not have a windows store prefs path either(something very rarely used or recommended within spicetify.) 
In my proposed system however, it would default to the normal spotify install prefs path if the windows store one was not detected.

ririxidev: I agree with charlie that we should at least merge it as temp fix until we'll find better way to handle it.

afonsojramos (approved):

PR #1584, opened 2022-04-12
fix: readme links

roythearsonist:
Links were broken so I fixed them.


afonsojramos (approved):
Thanks! Forgot about this!

PR #1587, opened 2022-04-14
fix(lyrics): switch to `wg://` protocol

### LINKED ISSUE #1566: Spotify Lyrics seems to be not working (Lyrics Plus and Popup)

- [X] I have checked the FAQ

```Markdown
- Spotify version: 1.1.81.604.gccacfc8c
- Spicetify version: 2.8.4
```

### 📝 Description

Even if Spotify is the first option on my Lyrics Plus, my client always retrieves Netease lyrics instead. I also tried disabling all the options but Spotify, but the result is as you see in the screenshots.
The only difference is that Popup Lyrics shows "No lyric".

### 📸 Screenshots

image
image
image

ririxidev:
* fix: change protocol to wg://
  * Fixes issues with getting spotify lyrics on lyrics-plus & popupLyrics
  * Adds @ts-nocheck on popupLyrics
* style: change wording when no lyrics

Fixes #1566

### DIFF
--- a/Extensions/popupLyrics.js
+++ b/Extensions/popupLyrics.js
@@ -1,4 +1,3 @@
-// @ts-nocheck
 
 // NAME: Popup Lyrics
 // AUTHOR: khanhas
--- a/Extensions/popupLyrics.js
+++ b/Extensions/popupLyrics.js
@@ -1,4 +1,3 @@
-
 // NAME: Popup Lyrics
 // AUTHOR: khanhas
 //         Netease API parser and UI from https://github.com/mantou132/Spotify-Lyrics
--- a/Extensions/popupLyrics.js
+++ b/Extensions/popupLyrics.js
@@ -87,7 +87,7 @@ function PopupLyrics() {
 
             const lines = body.lines;
             if (!lines || !lines.length || typeof lines[0].time !== "number") {
-                return { error: "No lyric" };
+                return { error: "No lyrics" };
             }
 
             const lyrics = lines.map((a) => ({
@@ -154,7 +154,7 @@ function PopupLyrics() {
                     }));
                     return { lyrics };
                 } else {
-                    return { error: "No lyric" };
+                    return { error: "No lyrics" };
                 }
             } catch (err) {
                 return { error: err.message };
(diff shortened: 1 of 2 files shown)


th3an7:
Hey,
How did you find about that new `wg://` protocol? I'm curious if there are any new endpoints or if they just moved old `hm://` without adding anything...

ririxidev:
> Hey,
> 
> How did you find about that new `wg://` protocol? I'm curious if there are any new endpoints or if they just moved old `hm://` without adding anything...

Someone said in linked issue that after changing to wg:// protocol it works, and it actually does so idk how they found it.

Extensions/popupLyrics.js
```diff
@@ -1,4 +1,4 @@
-// @ts-check
+// @ts-nocheck
```
afonsojramos, line 1:
```suggestion
// @ts-check
```

I'm not sure why you removed it, but you should get minimal errors with the check. Some are miscategorized, we are aware, but what you get from the `check` is more than what you'll get by just having it disabled.

afonsojramos (changes requested):
Looking good, just take a look at the requested changes. Sorry for the delay, Easter 🐰🥚 got in the way!

Extensions/popupLyrics.js
ririxidev: I would love to but it throws error that I can't fix since it's about types. This file isn't typescript one so I can't add types.

ririxidev: Having typescript check in JavaScript file isn't ideal and shouldn't be done at all because of errors. Typescript is for types, not for no having them and only using it for example "this return statement isn't going to work" etc etc. If you really need it I can simply add typescript file that will generate JavaScript dist

Extensions/popupLyrics.js
```diff
@@ -1,4 +1,3 @@
-// @ts-check
```
afonsojramos, line 1:
```suggestion
```

My bad here!

afonsojramos (changes requested):

afonsojramos (approved):

PR #1614, opened 2022-04-24
fix(devtools): overwrite homePath if snap detected

ririxidev:
I forgot to make checking if snapSpotifyHome exists and then overwrite homePath.


CharlieS1103 (approved):
That's mb, i missed it as well. Ty

PR #1613, opened 2022-04-24
fix(devtools): use new method to enable devtools

ririxidev:
This method is based on [PhilippIRL](https://gist.github.com/PhilippIRL/97908aba3a78cc0c8d0ab4e9439bf445)'s python script.
It 'patches' offline.bnk that is stored remote configuration, by changing some bytes we can achieve enabled devtools, but it resets every few time so you still need to write `spicetify enable-devtools` many times 😩 

Also changed `spicetify enable-devtool` to `spicetify enable-devtools`

PR #1618, opened 2022-04-25
fix: modify filePath from .config to .cache

Kafva:
Update the `--help` output in accordance with #1613


ririxidev:
If that's okay for you, I would rather include your fix in my pr that will fix enable-devtools command for linux users.

ririxidev (approved):

PR #1623, opened 2022-04-27
feat(restart): check if process is running

### LINKED ISSUE #1469: Only restart Spotify if it is already open
Restarting Spotify after running "spicetify restore backup apply" makes a lot of sense if it is already open, but right now it opens Spotify even if it's not. 

That doesn't make much sense. I think a check of it being open before "restarting" it should be implented.

I'm aware of the "-n" flag, but when it's actually open I want it to restart. I don't want it to start when I run "spicetify restore backup apply" when it's not though.

kyrie25:
Small QoL change, resolves #1469 
Spicetify should check whether process is running and would not launch it if it isn't
Tested and working for Windows
If `pgrep` output is accurately stated [here](https://linuxize.com/post/pgrep-command-in-linux/):
> The command returns 0 when at least one running process matches the requested name. Otherwise, the exit code is 1. This can be useful when writing shell scripts.

Then the result should be the same for Linux/MacOS

### DIFF
--- a/src/cmd/restart.go
+++ b/src/cmd/restart.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


src/cmd/restart.go
```diff
@@ -16,21 +17,33 @@ func RestartSpotify(flags ...string) {
...
 	case "windows":
-		exec.Command("taskkill", "/F", "/IM", "spotify.exe").Run()
-		if isAppX {
-			ps, _ := exec.LookPath("powershell.exe")
-			exe := filepath.Join(os.Getenv("LOCALAPPDATA"), "Microsoft", "WindowsApps", "Spotify.exe")
-			flags = append([]string{"-NoProfile", "-NonInteractive", `& "` + exe + `" --app-directory="` + appDestPath + `"`}, flags...)
-			exec.Command(ps, flags...).Start()
-		} else {
-			exec.Command(filepath.Join(spotifyPath, "spotify.exe"), flags...).Start()
+		isRunning := exec.Command("tasklist", "/FI", "ImageName eq spotify.exe")
```
ririxidev, line 20: image
?

src/cmd/restart.go
```diff
@@ -16,21 +17,33 @@ func RestartSpotify(flags ...string) {
...
+		_, err := isRunning.Output()
+		if err == nil {
+			exec.Command("pkill", "spotify").Run()
+			exec.Command(filepath.Join(spotifyPath, "spotify"), flags...).Start()
+		}
 	case "darwin":
-		exec.Command("pkill", "Spotify").Run()
-		flags = append([]string{"-a", "/Applications/Spotify.app"}, flags...)
-		exec.Command("open", flags...).Start()
+		isRunning := exec.Command("pgrep", "Spotify")
```
ririxidev, line 41: pgrep is bugged on macOS, try using `ps aux | grep "Spotify" | grep -v grep`

ririxidev (changes requested):

src/cmd/restart.go
kyrie25: Supposedly it works with the quotation marks (`tasklist /FI "ImageName eq Spotify.exe"`)
image
but will break if I add them in the file, works if it is without the quotes
image

src/cmd/restart.go
kyrie25: Just wanted a quick check, if I were to change it, would it be like so? I apologize for I am not familiar with the language
```suggestion
		isRunning := exec.Command("ps aux | grep Spotify | grep -v grep")
```
The command works on the terminal but currently I am unable to test with the executable for I don't have my hands on a Mac

ririxidev: Will check later on macOS.

kyrie25: Much appreciated!

kyrie25: After doing some research, the suggestion above is an invalid command and will return an error. I've found an alternative solution: 
```suggestion
		isRunning := exec.Command("grep", "Spotify")
		ps := exec.Command("ps", "aux")
		pipe, _ := ps.StdoutPipe()
		defer pipe.Close()
		isRunning.Stdin = pipe
		ps.Start()
```
It should be similar to `ps aux | grep "Spotify"` and should work as expected.

kyrie25: After doing some research I've found out that my suggestion above is an invalid command but I've managed to test my initial changes on a Mac and can confirm `pgrep` works as expected, so is there any chance you might reconsider it?
image

ririxidev: I said it's bugged out. Not that it doesn't work. You are on the latest macOS anyway, a lot of users use older versions. A lot of people complain that it doesn't work properly and recommend using brew's formula. We can't force people to install brew formula to have it working properly

kyrie25: Alright, understood. I'll try to find a better workaround for it.

kyrie25: Committed changes confirmed to be working on macOS
image

ririxidev (approved):

PR #1627, opened 2022-04-29
fix: Lyrics Plus not showing full lyrics

kyrie25:
Updates Regex to exclude the `<div>` element inside of `Lyrics__Container`
image
Tested and is working as expected
image
image


afonsojramos (approved):
Aaaaaawwwwesome PR!! 🥳 Ahaha Thanks for the fix!

samanft:
OMG TYSM!!

PR #1633, opened 2022-04-29
fix: adapt urls in custom app & extensions to `wg://` protocol

huhridge:
For `shuffle+`:
The new requests are just more reliable as compared to the `api-partner.spotify.com` requests.

The `fetchAlbum` one is just for sake of consistency.


ririxidev:
is that all what you want to change? if not, update this pr and lmk. For now not going to merge it since it's too much

huhridge:
@ririxidev I'm done now (i think)

Issue #1601, opened 2022-04-22
Keep getting `error Cannot backup app files. Reinstall Spotify and try again.` even after reinstalling spotify on fedora linux


ImDedInside-git:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version:
~ ❯ spotify -v                                    
Version: 1.1.80.699.gc3dac750_amd64.deb
Note: This is installed on Fedora Linux 35 with lpf [https://docs.fedoraproject.org/en-US/quick-docs/installing-spotify/]


- Spicetify version:
~ ❯ spicetify -v
2.9.5
```

### 📝 Description

Running `spicetify backup apply` gives this error.  
```Markdown
error Cannot backup app files. Reinstall Spotify and try again.
```

First time installing spicetify using curl commands 
```Markdown
curl -fsSL https://raw.githubusercontent.com/spicetify/spicetify-cli/master/install.sh | sh
curl -fsSL https://raw.githubusercontent.com/spicetify/spicetify-marketplace/main/install.sh | sh
```
Reinstalling spotify made no difference

### 📸 Screenshots

![image](https://user-images.githubusercontent.com/67794677/164720376-665a2986-88f4-4773-a443-8a7b80934935.png)

ririxidev:
spotify 1.1.80 isn't supported by v2.9.5.

theRealPadster:
You may need to change permissions on the folder that spotify is installed to

Lalit64:
Hey @ImDedInside-git I am getting the same error on macOS arm64 `spicetify backup apply enable-devtool`spicetify backup apply enable-devtool`

<img width="1686" alt="Screenshot 2022-04-29 at 12 02 21" src="https://user-images.githubusercontent.com/79628080/165924350-f4c08955-0942-4f51-9bf1-aa36089b7dd0.png">


And My Config:
<img width="1686" alt="Screenshot 2022-04-29 at 12 03 21" src="https://user-images.githubusercontent.com/79628080/165924501-d8bf7102-e6f7-46b5-b43a-3422797f1443.png">

xerta555:
> Permissions.

If this issue really come from permissions, i think it could be a good idea to specifying wich perms the folder and files must have

ririxidev:
Check your permissions, closing issue.

[...]

PR #1641, opened 2022-05-01
feat(backup): change error message

kyrie25:
Proposal PR, feel free to close if this is not appropriate.
- As #1601 points out that the [note](https://spicetify.app/docs/advanced-usage/installation#note-for-linux-users) sections can be neglected, thus checking if Spicetify have permissions to access app files _should be_ the first thing a user should check instead of reinstalling Spotify, which is relatively more of a hassle to do.
- Of course, this might _not_ be the common case and might be not that good to generalize it

### DIFF
--- a/src/cmd/backup.go
+++ b/src/cmd/backup.go
@@ -48,7 +48,7 @@ func Backup(spicetifyVersion string) {
 	if totalApp > 0 {
 		utils.PrintGreen("OK")
 	} else {
-		utils.PrintError("Cannot backup app files. Reinstall Spotify and try again.")
+		utils.PrintError("Cannot backup app files. Does Spicetify have permissions to access Spotify files?")
 		os.Exit(1)
 	}
 


src/cmd/backup.go
```diff
@@ -48,7 +48,7 @@ func Backup(spicetifyVersion string) {
 	if totalApp > 0 {
 		utils.PrintGreen("OK")
 	} else {
-		utils.PrintError("Cannot backup app files. Reinstall Spotify and try again.")
+		utils.PrintError("Cannot backup app files. Does Spicetify have permissions to access Spotify files?")
```
afonsojramos, line 51: So, unfortunately, there are more cases where this message occurs, so I do not feel like it matches all cases. Overall the current message targets most cases, therefore we should keep it for now.

kyrie25: Understood, thanks for your response.

PR #1647, opened 2022-05-04
fix(devtools): patch another location

ririxidev:
This commit should fix enabling devtools for some users. Now we will patch two locations in remote configuration cache to *just be sure*.

### DIFF
--- a/src/cmd/devtools.go
+++ b/src/cmd/devtools.go
@@ -45,9 +45,13 @@ func SetDevTools() {
 	buf := new(bytes.Buffer)
 	buf.ReadFrom(file)
 	content := buf.String()
-	location := strings.LastIndex(content, "app-developer")
-	patchLocation := int64(location + 15)
+	firstLocation := strings.Index(content, "app-developer")
+	firstPatchLocation := int64(firstLocation + 14)
 
-	file.WriteAt([]byte{50}, patchLocation)
+	secondLocation := strings.LastIndex(content, "app-developer")
+	secondPatchLocation := int64(secondLocation + 15)
+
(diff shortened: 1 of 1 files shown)


xerta555:
I just compile again `spicetify.exe` file and replace my one with the new one, the DevTools works again if i run `spicetify enable-devtools` command, as opposite as before when even if i did the command, it didn't works.
This new patch unfortunatly don't works if i restart Spotify normally.

Can we continue to investigate here or i should open a new issue?

CharlieS1103:
@xerta555 As far as I'm aware, the issue where it doesn't boot up again if you restart spotify normally is not solvable (at least not with significant difficulty)

xerta555:
> As far as I'm aware, the issue where it doesn't boot up again if you restart spotify normally is not solvable (at least not with significant difficulty)

Could you specify please?
I'm not sure to understand what you want to mean

CharlieS1103:
Pretty much, spotify automatically refreshes the offline.bnk file that we are modifying, we can't block Spotifys access to this file because it stores important information, therefore it would be intensely difficult to solve this issue.

xerta555:
> Pretty much, spotify automatically refreshes the offline.bnk file that we are modifying, we can't block Spotifys access to this file because it stores important information, therefore it would be intensely difficult to solve this issue.

I've just tested the method of changing the permissions of the file in question, and it doesn't work.
Even if i only allow R/W access to the file, do the command, then restart Spotify, the value doesn't stay at 2, it goes back to 0.
This means that another routine (in my case at least) comes and writes the value back to the file.

I have the impression that something else is set up differently on my side than on the other users for whom it works without any problem, butidk what exactly.

Thank you anyway.

PR #1701, opened 2022-05-21
feat: improve handling of new Spotify versions

afonsojramos:
We're always getting reports of the Cannot find index of React app. This fixes that and improves how we communicate to users that we only support until a certain release of Spotify. This also means that we should update the `supportedSpotifyVersion` every time that we release a new version to support a new Spotify version.

image

### DIFF
--- a/src/apply/apply.go
+++ b/src/apply/apply.go
@@ -18,6 +18,7 @@ type Flag struct {
 	HomeConfig    bool
 	ExpFeatures   bool
 	SpicetifyVer  string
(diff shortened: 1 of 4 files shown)


afonsojramos:
@ririxidev I get that, however, this is something for the CLI so... What do you suggest as an alternative? 

This PR also tackles the fact that currently the CLI breaks applying in the current version.

afonsojramos:
In the end, I just want to avoid this: 
image

PR #1775, opened 2022-06-24
fix Customization hyperlink

Sebight:
Point to development section rather to a page which results into 404.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,6 @@ Supports Windows, MacOS and Linux.
 ### Links
 - [Installation](https://spicetify.app/docs/getting-started)
 - [Basic Usage](https://spicetify.app/docs/getting-started#basic-usage)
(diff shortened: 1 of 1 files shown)


kyrie25:
Thanks for taking your time in making this PR!

However, I don't think the hyperlink's content quite matches its title so I'm not sure if it is good to merge. If you could find some other links from the docs that is more suitable with the title it'd be great!

Sebight:
Yes, I know. But it was still better than pointing to a page which is not existing. Maybe renaming the title or finding a better suiting content would do it.

kyrie25:
> Yes, I know. But it was still better than pointing to a page which is not existing.

I'd prefer to have a complete patch for the problem as to avoid bloating commit history in the future, so I ask that you understand my intention here.
Do let us know of any proposal that you have though!

PR #1783, opened 2022-06-26
chore(workflows): add commitlint

rxri:


kyrie25 (commented):
Should probably make this a required check also.

kyrie25 (approved):

afonsojramos:
Which rules does this proposed action follow?

kyrie25:
> Which rules does this proposed action follow?

Since a [config file isn't present](https://github.com/wagoid/commitlint-github-action#configfile), I believe it would follow the traditional [commitlint](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional)

PR #1820, opened 2022-07-14
Patch for certain flatpack installations

CharlieS1103:
Please note I did write this in github web code editor, had no way of testing this, and have never officially learned golang.


CharlieS1103:
General idea gets across, i'll fix it when i don't have a headache or someone else can do it lol.

PR #1825, opened 2022-07-19
fix(apply): custom app regex for `1.1.90`

kyrie25:
Compatibility with/hotfix for 1.1.90.

### DIFF
--- a/src/cmd/cmd.go
+++ b/src/cmd/cmd.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	supportedSpotifyVersion = "1.1.84"
+	supportedSpotifyVersion = "1.1.90"
 	spicetifyFolder         = utils.GetSpicetifyFolder()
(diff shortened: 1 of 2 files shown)


src/cmd/cmd.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	supportedSpotifyVersion = "1.1.84"
+	supportedSpotifyVersion = "1.1.90"
```
afonsojramos, line 16: Are you sure that these changes completely fix the error that occurs in [apply.go#L203](https://github.com/spicetify/spicetify-cli/blob/bb9d95ac1bbbb3d81ee8ec461e06f3240a439c62/src/apply/apply.go#L203)?

kyrie25: The error attached did not occur for me on this patch, nor did it occur for many versions now. So I assume that it's safe to bump.

afonsojramos (approved):

PR #1830, opened 2022-07-21
fix(css-map): add `main-avatar-image`

Tetrax-10:


kyrie25 (approved):
Looks about right, thanks for the help!

kyrie25:
> @ririxichan what is exactly the commitlint's output? I can't quite understand it.

If I may, it asks for the minimum requirement of a commitlint-compliant commit message
image

In this case, the commit message is missing both:
image

afonsojramos:
Oh... I see, you guys are evil, but congrats on adding that 😄

PR #1840, opened 2022-07-27
fix(shuffle+): skip unplayable tracks

iurimateus:
If a playlist contains an unavailable track, then Shuffle+ might fail to
add songs to the queue with: "PUT request to sp://player/v2/main/queue
failed with status code 403", resulting in an empty queue.

Similar to `fetchAlbum` and `fetchShow`, we should filter tracks that
are not playable.


Extensions/shuffle+.js
```diff
@@ -131,9 +131,10 @@
      */
     const fetchPlaylist = async (uri) => {
         const res = await Spicetify.CosmosAsync.get(`sp://core-playlist/v1/playlist/${uri}/rows`, {
-            policy: { link: true },
+            policy: { link: true, playable: true },
         });
-        return res.rows.map((item) => item.link);
+        const available = res.rows.filter((track) => track.playable);
+        return available.map((item) => item.link);
```
kyrie25, line 137: Don't really see the point in this.
```suggestion
        return res.rows.filter((track) => track.playable).map((item) => item.link);
```

iurimateus: Me neither, just keeping to what I've seen on https://github.com/spicetify/spicetify-cli/blob/master/Extensions/shuffle+.js#L224-L225. Fixed.

kyrie25 (approved):
Alright, tested and LGTM. Thanks for your contribution!

PR #1845, opened 2022-07-30
feat(fullAppDisplay): Add mouse binding to open config

vituskh:
For me, right clicking the Full App Display button does not work. This fixes that by setting f6 as a keybind for opening the config menu.

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -776,4 +776,6 @@ button.switch.disabled {
     );
 
     Spicetify.Mousetrap.bind("f11", toggleFad);
+    Spicetify.Mousetrap.bind("f6", openConfig);
+
 })();


kyrie25:
Hi! Thanks for taking your time in making this PR, but unfortunately I will have to turn this proposal down for the following reasons:

> For me, right clicking the Full App Display button does not work.

- This is because it is not meant to do so. Unlike other extensions such as `popupLyrics` where you can't have any interaction with the window, Full App Display config can be opened by **right-clicking** from within the Spotify client **while activating the extension**, and it should rightfully do so as the user can see live changes.

> This fixes that by setting f6 as a keybind for opening the config menu.

- It is preferred to have at least global keybinds as possible so as to not conflict with other existing extensions.

Hope you understand our decision and feel free to contribute in any other way that you see helpful :)

PR #1865, opened 2022-08-12
feat: reimplement `Experimental Features`

kyrie25:
Resolves #1485 (for real)
Tested and working for both Spotify `1.1.90` and `1.1.91`
Spotify_YVFlpqsAGv

<details>
<summary>Technical stuff</summary>

Since `RemoteConfigResolver` is no longer available from `1.1.78`, raw data is fetched from `vendor~xpui` and modified in there instead, using only one method to fetch and override instead of two like before.

Also because there is only one method to fetch these features, there is no effective way to cross-reference and see if old experimental features are still available in newer versions, thus why a `Reset` button is added.

image

</details>

PR #1867, opened 2022-08-12
feat(expFeatures): update new features upon mismatch

kyrie25:
~~Everything is done, for real this time~~
Resolves several inconveniences mentioned in #1865 such as:
~~- Feature list will now automatically updates upon Spotify version mismatch (the `Reset` button stays tho, just in case)
Will now have a notice on the top that tells users to reload for the new changes to apply.~~
Apparently I don't even need this anymore, it just purge the old ones now.

image

- List will not be shown if hooks are not patched/hasn't loaded yet, which will remove the need for the placeholder text I added before (again, it will stay in case it *does* break again, at least we will know what broke it)
https://github.com/spicetify/spicetify-cli/blob/cf1598dd66d3a27875e01c67369f742d291bc8d3/jsHelper/expFeatures.js#L189

### DIFF
--- a/jsHelper/expFeatures.js
+++ b/jsHelper/expFeatures.js
@@ -1,5 +1,7 @@
 (function () {
-    let overrideList;
+    let overrideList,
+        newFeatures,
+        hooksPatched = false;
     try {
         overrideList = JSON.parse(localStorage.getItem("spicetify-exp-features"));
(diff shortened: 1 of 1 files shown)


ririxichan (approved):

kyrie25:
Need to refactor something first before merging.

theRealPadster (approved):

PR #1882, opened 2022-08-23
feat(popup-lyrics): add global delay option

Woosy:
Hi, here is my first PR, which adds an option to the Popup Lyrics settings menu to set a global delay.

I implemented this because when I'm listening to music with my Airpods on my iPhone, while working on my computer, there always is a delay between the lyrics and the song, and I have to press pause/resume every time to fix it. With this new option, it is now possible to avoid this problem.

By default, delay is set to 0. I made the choice to use an `input type="number"` because a slider / a select / two minus and plus buttons, were not convenient to select a precise value.

### DIFF
--- a/Extensions/popupLyrics.js
+++ b/Extensions/popupLyrics.js
@@ -261,6 +261,7 @@ function PopupLyrics() {
         blurSize: Number(LocalStorage.get("popup-lyrics:blur-size")),
         fontFamily: LocalStorage.get("popup-lyrics:font-family") || "spotify-circular",
         ratio: LocalStorage.get("popup-lyrics:ratio") || "11",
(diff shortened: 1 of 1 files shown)


ririxiiscute (changes requested):
Please run prettier to prettify file

kyrie25:
Hi! Thanks for your contribution.
Can you provide screenshots showing the new changes?

Woosy:
> Hi! Thanks for your contribution. Can you provide screenshots showing the new changes?

Here is a screenshot where you can see the new option:
image

ririxiiscute (approved):

kyrie25 (approved):
LGTM! Thanks for the work :)

PR #1909, opened 2022-09-01
fix(css-map): add main-entityHeader-imagePlaceholder

Tetrax-10:
Image Holder class eg: behind local files folder image in your library

### DIFF
--- a/css-map.json
+++ b/css-map.json
@@ -1858,5 +1858,6 @@
     "SkbGMKYv49KtJNB5XxdX": "x-settings-input",
     "uIJTvxFOg2izOY7aRRiU": "main-home-home",
     "I3EivnXTjYMpSbPUiYEg": "main-home-content",
-    "tlNAdRNP5lIeiD85nMcL": "main-entityHeader-imagePlaceholder"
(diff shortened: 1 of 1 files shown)


kyrie25:
Can you take a screenshot of where these classes are mapped? Local Files playlist doesn't have an image...?
image
`main-editImageButton-overlay` is already mapped in latest version of Spotify (`1.1.93`):
image
`main-entityHeader-imagePlaceholder` is already mapped in latest version of Spotify (`1.1.93`):
image

Tetrax-10:
I fixed [My theme](https://github.com/Tetrax-10/Nord-Spotify/commit/1b3a35d56a7215846cd1b554df00d11fea005756) with these classes for 1.1.93 they worked. But I didn't check with older version. I will check and reopen this PR

kyrie25:
> I fixed [My theme](https://github.com/Tetrax-10/Nord-Spotify/commit/1b3a35d56a7215846cd1b554df00d11fea005756) with these classes for 1.1.93 they worked. But I didn't check with older version. I will check and reopen this PR

I've noticed that these classes are used but since there are no equivalent of those classes mapped, I'd recommend *using* the mapped classes as a referrer in themes, eg. `.main-editImageButton-overlay > div` so that it would not break so often.

PR #1921, opened 2022-09-08
fix(shuffle+): improved and fixed bugs

Tetrax-10:
1.Added support for Singles, artist liked songs, top 10 songs, ep, in artist page
image
image

2.Improved queue
3.Fixed liked songs unplayable tracks

4.Removed auto shuffle as it breaks internal stuff and also causes issues in episodes page and other unsupported pages
shuffle+ error with auto play


Extensions/shuffle+.js
```diff
@@ -1,148 +1,277 @@
...
+                },
+                onclickFun: async () => {
+                    // console.log(CONFIG.artistMode);
+                },
+            }),
+            React.createElement(checkBoxItem, {
+                name: "Chosen artist must be included",
+                field: "artistNameMust",
+                onclickFun: async () => {
+                    // console.log(CONFIG.artistNameMust);
```
kyrie25, line 203: These `onclick` functions essentially do nothing?

kyrie25 (changes requested):

Extensions/shuffle+.js
Tetrax-10: grammar? ok fine

Extensions/shuffle+.js
```diff
@@ -1,148 +1,269 @@
...
-                    case Spicetify.URI.Type.SHOW:
-                    case Spicetify.URI.Type.PLAYLIST:
-                    case Spicetify.URI.Type.PLAYLIST_V2:
-                    case Spicetify.URI.Type.FOLDER:
-                    case Spicetify.URI.Type.ALBUM:
-                    case Spicetify.URI.Type.COLLECTION:
-                    case Spicetify.URI.Type.ARTIST:
-                        return true;
+    const CONFIG = await getConfig();
+    await saveConfig();
```
kyrie25, line 43: Is it required to do this on startup?

Extensions/shuffle+.js
```diff
@@ -1,148 +1,269 @@
...
+                    "div",
+                    { className: "col action" },
+                    React.createElement(
+                        "button",
+                        {
+                            className: "checkbox" + (value ? "" : " disabled"),
+                            onClick: async () => {
+                                let state = !value;
+                                CONFIG[field] = state;
+                                setValue(state);
```
kyrie25, line 132:
```suggestion
                                CONFIG[field] = !value;
                                setValue(!value);
```
Don't really need a var for this since you've done the same without on other functions also

Extensions/shuffle+.js
```diff
@@ -1,148 +1,269 @@
...
+            React.createElement("div", { className: "popup-row" }, React.createElement("h3", { className: "div-title" }, "Artist Shuffle")),
+            React.createElement("div", { className: "popup-row" }, React.createElement("hr", { className: "divider" }, null)),
+            React.createElement(dropDownItem, {
+                name: "Shuffle mode Artist Page",
+                field: "artistMode",
+                options: {
+                    all: "All",
+                    album: "Albums",
+                    single: "Singles & EP",
+                    likedSongArtist: "Liked Songs of Artist",
```
kyrie25, line 190:
```suggestion
                    likedSongArtist: "Artist's Liked Songs",
```

Extensions/shuffle+.js
```diff
@@ -335,13 +464,97 @@
...
+                if (CONFIG.artistMode == "topTen") {
+                    Spicetify.showNotification(`Shuffled Top ${count} Songs`);
+                    break;
+                }
+                if (CONFIG.artistMode == "likedSongArtist") {
+                    Spicetify.showNotification(`Shuffled ${count} Liked Songs`);
+                    break;
+                }
+                if (CONFIG.artistMode == "single") {
+                    Spicetify.showNotification(`Shuffled ${artistFetchTypeCount.single} Singles, Totally ${count} Songs`);
```
kyrie25, line 480:
```suggestion
                    Spicetify.showNotification(`Shuffled ${artistFetchTypeCount.single} Singles, Total of ${count} Songs`);
```
Same for the cases below.

Extensions/shuffle+.js
```diff
@@ -1,148 +1,269 @@
...
 
-(function ShufflePlus() {
-    if (!Spicetify.CosmosAsync || !Spicetify.Platform) {
-        setTimeout(ShufflePlus, 1000);
+let shufflePlusCount = 0;
+(async function shufflePlus() {
+    if (!(Spicetify.CosmosAsync && Spicetify.Platform && shufflePlusCount < 200)) {
+        setTimeout(shufflePlus, 300);
+        shufflePlusCount++;
         return;
```
kyrie25, line 12: Why do you need the function to recursively run 200 times before it is executed?

kyrie25 (changes requested):
Take a look into these also.

Extensions/shuffle+.js
Tetrax-10: yes

Extensions/shuffle+.js
Tetrax-10: ok

Extensions/shuffle+.js
Tetrax-10: no

Extensions/shuffle+.js
Tetrax-10: mo

Extensions/shuffle+.js
Tetrax-10: this is how ext works

Extensions/shuffle+.js
kyrie25: Grammatically incorrect. Please do change it.

Extensions/shuffle+.js
kyrie25: > yes

Why so? The config isn't changed until the menu item is appended much later on.

kyrie25: > dont remove
> 
> ```js
> case Spicetify.URI.Type.ARTIST:
>                         return true;
> ```

I wasn't referring to it.

Extensions/shuffle+.js
kyrie25: > it does't run for 200 times. 200 is the limit, prob spicetify will be available with in 3 to 10 loops. if spicetify is not available then this function will loop infinite no of time. so i set a limit 200

If both `Spicetify.Platform` and `Spicetify.CosmosAsync` is not available, this is a major CLI bug that needs to be patched. This is not required.

kyrie25: > it does't run for 200 times. 200 is the limit.

The function would continue to execute as normal when the counter reaches 200 even when both `Platform` and `CosmosAsync` are not available, which breaks the extension anyway and defeats the purpose of the null check
image

Extensions/shuffle+.js
Tetrax-10: > > yes
> 
> Why so? The config isn't changed until the menu item is appended much later on.

it is the init of the config like default settings will be stored. if you remove it the places with `CONFIG.something` wont work

Extensions/shuffle+.js
kyrie25: > this is my way of writing if you don't like it. you change whatever you want am not gonna reject any changes from now. i have my version with me, so its up to you.

It's not a matter of style, this is actually crucial to the extension. I'm not the antagonist here nor am I trying to nitpick or reject your work, these are constructive suggestions with explanations attached that helps you understand my decision on why I commented on it. Please do understand that.

Extensions/shuffle+.js
kyrie25: Understandable. Thank you.

Extensions/shuffle+.js
kyrie25: I do apologize if I gave such impression and hope that you understand my intentions here.

Tetrax-10: > 

fixed it

Tetrax-10: 6 seconds ?? 300x200 = 60 seconds

```js
let shufflePlusCount = 0;
(async function shufflePlus() {
    if (!(Spicetify.CosmosAsync && Spicetify.Platform) && shufflePlusCount < 200) {
        setTimeout(shufflePlus, 300);
        shufflePlusCount++;
       console.log(shufflePlusCount); // add this line and run
        return;
    }
    await initShufflePlus();
})();
```

Tetrax-10: This counter is just an extra layer to prevent bugs nothing wrong in that.

For example in `Skip or Play Liked Songs` extension if no audio device is connected the extension will loop infinite times because `Platform ` and `LocalStorage` will be available but not `Player`. so 60 sec is the limit if no audio device is connected then it will stop init process
image

here for eg lets assume `CosmosAsync` hook broke i dont want this init process to run infinite times. so gave 60 sec time limit

kyrie25: > This counter is just an extra layer to prevent bugs nothing wrong in that.

Shuffle+ would not execute with either of those hooks broke. The bugs wouldn't even be found unless the CLI is patched as mentioned before.

> here for eg lets assume `CosmosAsync` hook broke i dont want this init process to run infinite times. so gave 60 sec time limit.

If either of those hooks broke, none of Spicetify functionality (except for themes) would work, and stopping the recursive initialization would not bring any significant benefit except for adding another unnecessary and confusing condition to those who review the extension.

> 6 seconds ?? 300x200 = 60 seconds

My bad, point still stands.

kyrie25 (approved):

Extensions/shuffle+.js
```diff
@@ -1,148 +1,266 @@
...
-                    case Spicetify.URI.Type.ALBUM:
-                    case Spicetify.URI.Type.COLLECTION:
-                    case Spicetify.URI.Type.ARTIST:
-                        return true;
+    const CONFIG = await getConfig();
+    await saveConfig();
+
+    async function saveConfig() {
+        await setLocalStorageDataWithKey("shufflePlus:settings", JSON.stringify(CONFIG));
+    }
```
afonsojramos, line 45: Overall I don't like wrapping one liners, but I think here you made a good decision.

afonsojramos (approved):

PR #1940, opened 2022-09-18
fix: handling usernames with unicode characters

kyrie25:
This error is exclusive to Windows user afaik
Because [Powershell has a *very* bad way of printing out Unicode characters and still doesn't have full support for it](https://devblogs.microsoft.com/commandline/windows-command-line-inside-the-windows-console/#handling-unicode), using their internal functions to manipulate strings with those characters will print out gibberish that will throw errors on Marketplace install scripts or anything that relies on `spicetify -c | Split-Path` 
Screenshot is using latest stable version of PS, showing the error still persists to this day
image

Running the current install script will give out such ([taken by person with the errors themselves](https://discord.com/channels/842219447716151306/1015684719712686150/1015894752517967932)):
image

This should provide a way to print out config folder path without having to rely on Powershell functions

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -139,7 +139,11 @@ func main() {
 		return
 
 	case "config-dir":
-		cmd.ShowConfigDirectory()
+		if quiet {
+			log.SetOutput(os.Stderr)
+			os.Stdout = os.Stderr
+		}
+		cmd.ShowConfigDirectory(quiet)
 		return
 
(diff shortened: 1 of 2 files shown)


iamririxi (approved):

kyrie25:
In hindsight, since config path is immutable, it might be better to just patch the install script with hard-coded values.

PR #1941, opened 2022-09-18
feat(expFeatures): enable usage for enum type

kyrie25:
Enables user the choice to change Experimental features that uses enums in addition to boolean types.
Spotify_RJfFkYKDfM


iamririxi (approved):

theRealPadster (approved):
👍

PR #1944, opened 2022-09-20
fix(fad): disable toggle if `Lyrics Plus` not applied

kyrie25:
so people would stop asking the same question on support channel
image

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -686,11 +686,15 @@ body.video-full-screen.video-full-screen--hide-ui {
         style.innerHTML =
             styleBase +
             styleChoices[CONFIG.vertical ? 1 : 0] +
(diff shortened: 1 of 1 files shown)


CharlieS1103 (commented):
Perhaps instead of setting the name to Unable to find lyrics plus, it'd be more clear to have a pop up notification that explicitly states that the user needs to install lyrics plus after the user enables the feature

kyrie25:
> Perhaps instead of setting the name to Unable to find lyrics plus, it'd be more clear to have a pop up notification that explicitly states that the user needs to install lyrics plus after the user enables the feature

`showNotification` doesn't display over fad, plus they might not notice it and then asks the question away regardless, which brings us back to people spamming questions on support server.

theRealPadster (approved):

iamririxi (approved):

PR #1960, opened 2022-09-28
fix: clear path if invalid AppX path is used

kyrie25:
- Reset path if `prefs_path` is of AppX and invalid (which is the majority case if `prefs_path` issue on Windows), hopefully this solves a handful of issues like this (if not all). This should also ease up the switching process from AppX to normal version after the user reads the message mentioned below:
Code_GLV8UpIaOJ
Also found out that the AppX `spotify_path` may still exist after an uninstallation, so added an extra check to be sure.

- Shows MS Store version warn message on backup along with the `Spicetify not apply on restart` solution and forces the user to respond (and hopefully read) before continuing:
Code_dNty039Xrw

### DIFF
--- a/src/cmd/backup.go
+++ b/src/cmd/backup.go
@@ -16,6 +16,14 @@ import (
 // Backup stores original apps packages, extracts them and preprocesses
 // extracted apps' assets
 func Backup(spicetifyVersion string) {
+	if isAppX {
+		utils.PrintInfo(`You are using Spotify Windows Store version, which is only partly supported.
+Stop using Spicetify with Windows Store version unless you absolutely CANNOT install normal Spotify from installer.
+Modded Spotify cannot be launched using original Shortcut/Start menu tile. To correctly launch Spotify with modification, please make a desktop shortcut that execute "spicetify auto". After that, you can change its icon, pin to start menu or put in startup folder.`)
+		if !ReadAnswer("Continue backing up anyway? [y/N]: ", false, true) {
+			os.Exit(1)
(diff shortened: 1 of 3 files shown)


src/cmd/backup.go
```diff
@@ -16,6 +16,14 @@ import (
...
 // extracted apps' assets
 func Backup(spicetifyVersion string) {
+	if isAppX {
+		utils.PrintInfo(`You are using Spotify Windows Store version, which is only partly supported.
+Stop using Spicetify with Windows Store version unless you absolutely CANNOT install normal Spotify from installer.
+Modded Spotify cannot be launched using original Shortcut/Start menu tile. To correctly launch Spotify with modification, please make a desktop shortcut that execute "spicetify auto". After that, you can change its icon, pin to start menu or put in startup folder.`)
+		if !ReadAnswer("Continue backing up anyway? [y/N]: ", false, true) {
+			os.Exit(1)
+		}
+	}
```
iamririxi, line 26: You added this in other pr and I think it can make an issue while merging this one but idk

iamririxi (changes requested):

src/cmd/backup.go
kyrie25: Yeah I froze for a bit when I found it lmao, I'll merge it in this one instead since it comes with the whole package.

iamririxi (approved):

afonsojramos (approved):
Great polish!! ✨

PR #1965, opened 2022-09-30
feat(expFeatures): general overhaul

kyrie25:
Works on `1.1.92` and later.
`expFeatures` now override active properties and apply changes in real-time instead of just replacing the default value locally. This also means that **remotely enabled/disabled features** can now be overwritten, and the client no longer needs to reload every time the modal is closed:
Spotify_2iv85IQL2B
Add search modal:
Spotify_0x4euF0OuZ

### DIFF
--- a/jsHelper/expFeatures.js
+++ b/jsHelper/expFeatures.js
@@ -1,14 +1,21 @@
 (function () {
     let overrideList,
         newFeatures = [],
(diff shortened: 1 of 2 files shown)


iamririxi (approved):

afonsojramos (approved):
Amazing work 🙏 🙌

PR #1976, opened 2022-10-08
fix(css-map): map `Search` classes

kyrie25:


iamririxi (approved):

Tetrax-10:
> _ No description provided. _

How do you name these classes? you name them as you wish or they are in registered in some internal js files?

afonsojramos (approved):

afonsojramos:
> > _ No description provided. _
> 
> How do you name these classes? you name them as you wish or they are in registered in some internal js files?

Yep, pretty much, but we don't exactly have a good strong alternative... We're chained by Spotify's will 🤷

PR #1985, opened 2022-10-15
fix(install.sh): create shellrc file when not found

theRealPadster:
For install.sh, this will create a blank shellrc file if it doesn't exist. This should fix it so that the install succeeds when the user has no shellrc file. 

As a side note, `$SHELL` doesn't work in Fish. It returns `/bin/zsh` on macOS when running Fish.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,6 +65,14 @@ EOINFO
 check() {
 	local path="export PATH=\$PATH:$spicetify_install"
 	local shellrc=$HOME/$1
+
+	# Create shellrc if it doesn't exist
+	if ! [ -f $shellrc ]; then
+		echo "CREATING $shellrc"
+		touch $shellrc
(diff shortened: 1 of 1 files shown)


iamririxi:
There is no need for that tho. If you do `>>` it creates file as well if it doesn't exist
image

iamririxi:
nvm didn't understand this lol

iamririxi (approved):

CharlieS1103:
Just fix lint stuff and should be good

theRealPadster:
> Just fix lint stuff and should be good

I don't see what the problem is? 
image

theRealPadster:
Yeah I think it's that last commit name. Just squash it and name it whatever you want...

CharlieS1103 (approved):

PR #1986, opened 2022-10-17
feat: add Spicetify config to version modal

kyrie25:
Utilize `Spicetify.Config` by adding it to the version modal for ease of troubleshooting. Pressing the copy button will copy everything visible inside the modal.
Since most users are using a hybrid of Marketplace/CLI extensions, this should also be a more complete way of getting real-time config thus should help.
image
Though I noticed that if the `details` element is not opened, the copy button won't fetch the `li` elements inside so should we make it expand by default?


theRealPadster:
This is a good idea! Does it copy the Spicetify version if it's closed? That's probably good, and then we can tell people to expand it to get the rest?

theRealPadster (approved):

iamririxi (approved):

kyrie25:
> This is a good idea! Does it copy the Spicetify version if it's closed? That's probably good, and then we can tell people to expand it to get the rest? 

Yeah it does, just the contents inside aren't copied until expanded (Themes, Exts, Custom Apps).

iamririxi:
> > This is a good idea! Does it copy the Spicetify version if it's closed? That's probably good, and then we can tell people to expand it to get the rest?
> 
> Yeah it does, just the contents inside aren't copied until expanded (Themes, Exts, Custom Apps).

Seems reasonable for me

PR #1990, opened 2022-10-22
feat(lyrics-plus): add japanese lyrics conversion tool

41pha1:
This adds the functionality of switching between different display modes for Japanese lyrics.


theRealPadster:
Probably should have the dictionaries included in the same repo

41pha1:
> Probably should have the dictionaries included in the same repo

Alright I included them in the PR now

CharlieS1103:
@theRealPadster You sure? Seems like it would add quite a bit of bloat to the install, although I didn't really check.

41pha1:
Loading them dynamically using jsdeliver seems to work just fine

CustomApps/lyrics-plus/index.js
```diff
@@ -464,14 +510,14 @@ class LyricsContainer extends react.Component {
...
-                    lyrics: this.state.synced,
+                    lyrics: CONFIG.visual["translate"] && translatedLyrics ? translatedLyrics : this.state.synced,
                     provider: this.state.provider,
                     copyright: this.state.copyright,
                 });
             } else if (mode === UNSYNCED && this.state.unsynced) {
                 activeItem = react.createElement(UnsyncedLyricsPage, {
                     trackUri: this.state.uri,
-                    lyrics: this.state.unsynced,
+                    lyrics: CONFIG.visual["translate"] && translatedLyrics ? translatedLyrics : this.state.synced,
```
ririxiqt, line 520: shouldn't it be
```suggestion
                    lyrics: CONFIG.visual["translate"] && translatedLyrics ? translatedLyrics : this.state.unsynced,
```
since it's about unsyced lyrics, not synced ones

CustomApps/lyrics-plus/manifest.json
```diff
@@ -76,6 +76,9 @@
         "OptionsMenu.js",
         "TabBar.js",
         "Utils.js",
-        "Settings.js"
+        "Settings.js",
+        "Translator.js",
+        "dict_part_1.js",
+        "dict_part_2.js"
```
ririxiqt, line 82: well... i would rather load these dict files when it's needed. 30mb is a lot, since it adds another megabytes to download for people, even when they don't want it

ririxiqt (changes requested):

ririxiqt:
lmao I can't type xD ^

CustomApps/lyrics-plus/manifest.json
41pha1: Yes that's quite understandable, how about making it so that there's a download button when first opening the conversion menu on a Japanese song? The dicts will be cached locally after that.

CustomApps/lyrics-plus/README.md
```diff
@@ -22,6 +22,10 @@ Lyrics in Unsynced and Genius modes can be search and jump to. Hit Ctrl + Shift
 
 ![search](./search.png)
 
+Choose between different option of displaying japanese lyrics. [Furigana, Romaji, Hirgana, Katakana]
```
kyrie25, line 25:
```suggestion
Choose between different option of displaying Japanese lyrics. (Furigana, Romaji, Hirgana, Katakana)
```

CustomApps/lyrics-plus/README.md
```diff
@@ -35,3 +39,4 @@ spicetify apply
 
 -   A few parts of app code are taken from Spotify official app, including SyncedLyricsPage, CSS animation and TabBar. Please do not distribute these code else where out of Spotify/Spicetify context.
 -   Netease synced lyrics parser is adapted from [mantou132/Spotify-Lyrics](https://github.com/mantou132/Spotify-Lyrics). Give it a Star if you like this app.
+-   The Algorithm for converting japanese lyrics is based on [Hexenq's Kuroshiro](https://github.com/hexenq/kuroshiro).
```
kyrie25, line 42:
```suggestion
-   The algorithm for converting Japanese lyrics is based on [Hexenq's Kuroshiro](https://github.com/hexenq/kuroshiro).
```

CustomApps/lyrics-plus/index.js
```diff
@@ -220,6 +227,7 @@ class LyricsContainer extends react.Component {
     }
 
     async fetchLyrics(track, mode = -1) {
+        this.state.furigana, (this.state.romaji = null);
```
kyrie25, line 230: What does this one-liner do exactly? Did you mean to assign both of them the value `null`?

kyrie25 (changes requested):
This is really amazing work! 😄 Thanks a lot for your contribution.
I just have a few small suggestions regarding this PR.

CustomApps/lyrics-plus/manifest.json
kyrie25: Since both the dictionary files are already inside the custom app folder, this means that it will be included in CLI releases and users would only need to download them whenever they upgrade the CLI. Sounds reasonable to me tbh, but it still means that the CLI package is now 10 times larger than it ever was:
image

If we don't want to do this, then the only option left is to *not* include it into CLI releases and use it as a remote script via `jsdelivr` like before. This would mean that users can prompt when to download the dictionary, but the client would need to do it every time cache is dumped.

CharlieS1103: To be entirely honest I think this is the best solution, multiplying the cli size by 10 does not seem worth it to me. Particularly since a majority of people probably won't ever do so much as touch this feature. Or, perhaps we could merge the cli changes but make the modified lyrics plus it's own separate repo. (Kind of similar to when we made it so custom app owners could run code like an extension for marketplace, instead of building marketplace into the cli repo like expFeatures or home/sidebar config.)

CharlieS1103: Ah, just checked back and saw that there are actually no cli changes, why not just make this a separate repo altogether?

CharlieS1103:
How about we make it so lyrics plus uses his translation or code or whatever initially with just hotloading but if the user installs the dicts locally via an install script then it uses that method instead? So users who don't need the feature don't need the extra 30mb and the users who often use the feature and don't want the load times can install it

CustomApps/lyrics-plus/index.js
41pha1:
```suggestion
        this.state.furigana = this.state.romaji = this.state.hirgana = this.state.katakana = null;
```

kyrie25: Yep that sounds about right, commit whenever you see fit.

41pha1:
The dictionaries will now be dynamically loaded if they are not yet present and the conversion menu will show a loading indicator. 

If the dicts are somehow added to the client in any other way, it will use them instead and skip the downloading step.

(I havent come up with a better place than my [private repo](https://github.com/41pha1/spicetify-romaji-lyrics/tree/main/marketplace) yet)

CustomApps/lyrics-plus/Translator.js
```diff
@@ -0,0 +1,6102 @@
...
+                            P = /^(0x[0-9A-F]{4})\.\.(0x[0-9A-F]{4})(?:\s+([^#\s]+))(?:\s+([^#\s]+))*/;
+                        function h() {
+                            (this.char_def = new v()),
+                                (this.char_def.invoke_definition_map = new j()),
+                                (this.character_category_definition = []),
+                                (this.category_mapping = []);
+                        }
+                        (h.prototype.putLine = function (f) {
+                            var n = x.exec(f);
+
```
41pha1, line 6051: This should probably change

CustomApps/lyrics-plus/index.js
```diff
@@ -275,10 +275,13 @@ class LyricsContainer extends react.Component {
             return;
         }
 
-        if (!this.state.synced) return;
+        let lyricsToTranslate = this.state.synced;
+
+        if (!lyricsToTranslate) lyricsToTranslate = this.state.unsynced;
+        if (!lyricsToTranslate) return;
```
kyrie25, line 281:
```suggestion
        let lyricsToTranslate = this.state.synced;

        if (!lyricsToTranslate) {
        	lyricsToTranslate = this.state.unsynced;
        	return;
        }
```
Also is the variable reassigned anywhere? If not then use `const` instead of `let`.

ririxidev: well, no it doesn't make sense, it should be
```suggestion
        let lyricsToTranslate = this.state.synced ?? this.state.unsynced;

        if (!lyricsToTranslate) return;
```

41pha1: > well, no it doesn't make sense, it should be

Yea this seems to clear it up a bit!

CustomApps/lyrics-plus/Utils.js
```diff
@@ -54,4 +54,35 @@ class Utils {
...
+            )
+                return true;
+        return false;
+    }
+
+    static rubyTextToReact(s) {
+        const react = Spicetify.React;
+
+        let rubyElems = s.split("<ruby>");
+        let reactChildren = [];
```
ririxidev, line 73:
```suggestion
        const rubyElems = s.split("<ruby>");
        const reactChildren = [];
```

CustomApps/lyrics-plus/index.js
```diff
@@ -236,24 +244,63 @@ class LyricsContainer extends react.Component {
...
+        for (let lyric of lyricsToTranslate) lyricText += lyric.text + "\n";
+
+        [
+            ["romaji", "spaced", "romaji"],
+            ["hiragana", "furigana", "furigana"],
+            ["hiragana", "normal", "hiragana"],
+            ["katakana", "normal", "katakana"],
+        ].map((params) =>
+            this.translator.romajifyText(lyricText, params[0], params[1]).then((result) => {
+                let translatedLines = result.split("\n");
```
ririxidev, line 292:
```suggestion
                const translatedLines = result.split("\n");
```

CustomApps/lyrics-plus/index.js
```diff
@@ -450,6 +497,7 @@ class LyricsContainer extends react.Component {
             }
         }
 
+        let translatedLyrics = this.state[CONFIG.visual["translation-mode"]];
```
ririxidev, line 500:
```suggestion
        const translatedLyrics = this.state[CONFIG.visual["translation-mode"]];
```

CustomApps/lyrics-plus/index.js
```diff
@@ -509,6 +557,8 @@ class LyricsContainer extends react.Component {
         }
 
         this.state.mode = mode;
+        let showTranslationButton = (this.state.synced || this.state.unsynced) && Utils.isJapanese(this.state.synced || this.state.unsynced) && (mode == SYNCED || mode == UNSYNCED);
+        let translatorLoaded = this.translator.finished;
```
ririxidev, line 561:
```suggestion
        const showTranslationButton = (this.state.synced || this.state.unsynced) && Utils.isJapanese(this.state.synced || this.state.unsynced) && (mode == SYNCED || mode == UNSYNCED);
        const translatorLoaded = this.translator.finished;
```

ririxidev (changes requested):
Please run https://prettier.io  on all javascript files as well

CustomApps/lyrics-plus/Utils.js
```diff
@@ -54,4 +54,35 @@ class Utils {
...
+        const react = Spicetify.React;
+
+        const rubyElems = s.split("<ruby>");
+        const reactChildren = [];
+
+        reactChildren.push(rubyElems[0]);
+
+        for (let i = 1; i < rubyElems.length; i++) {
+            let kanji = rubyElems[i].split("<rp>")[0];
+            let furigana = rubyElems[i].split("<rt>")[1].split("</rt>")[0];
```
ririxidev, line 79:
```suggestion
            const kanji = rubyElems[i].split("<rp>")[0];
            const furigana = rubyElems[i].split("<rt>")[1].split("</rt>")[0];
```

I didn't notice it at first glance and these variables aren't re-assigned so they can be assigned as `const`

ririxidev (changes requested):
That's all I noticed.
About dict files, we'll see if we'll keep them in different repository instead of your own

CustomApps/lyrics-plus/Utils.js
41pha1: > That's all I noticed. About dict files, we'll see if we'll keep them in different repository instead of your own

Thats sound just about perfect!  Thanks for all of the suggestions, ill make sure to use const more often in the future xD

ririxidev (commented):
Please resolve conflicts that appeared after merging #1992. We created an repo for dict files - https://github.com/spicetify/spicetify-lyrics-romaji. Is it okay with you if we will keep these dict files in this repo instead?

41pha1:
> Please resolve conflicts that appeared after merging #1992. We created an repo for dict files - https://github.com/spicetify/spicetify-lyrics-romaji. Is it okay with you if we will keep these dict files in this repo instead?

That seems like the perfect place for them! Could you add the dict files? I sadly cant seem to open a PR on an empty repo.
Also prettier seems to be erroring after merging, looks like I might have messed something up?

ririxidev:
Don't worry about Prettier Action. For some reason it doesn't use your fork but instead this repo.
I will upload these files when I will be able to and will let you know

ririxidev:
#2000 probably fixed prettier action in this pull request. Additionally, I uploaded dictionary files to [spicetify-lyrics-romaji](https://github.com/spicetify/spicetify-lyrics-romaji) repo which you can find [here](https://github.com/spicetify/spicetify-lyrics-romaji/tree/main/dictionary)

41pha1:
> #2000 probably fixed prettier action in this pull request. Additionally, I uploaded dictionary files to [spicetify-lyrics-romaji](https://github.com/spicetify/spicetify-lyrics-romaji) repo which you can find [here](https://github.com/spicetify/spicetify-lyrics-romaji/tree/main/dictionary)

Awesome thanks ^^ I changed the urls to now point to the new location and it seems to work!

ririxidev (approved):
i need to look into prettier action again :/

kyrie25 (approved):
LGTM

huhridge:
I tested this, and it works great! I was gonna add this, but this is better than I would have implemented it.

Is there a possibility of adding a check to only translate when the song is actually Japanese? Right now, it passes everything, and it messes with the formatting of the lyrics.

41pha1:
> I tested this, and it works great! I was gonna add this, but this is better than I would have implemented it.
> 
> Is there a possibility of adding a check to only translate when the song is actually Japanese? Right now, it passes everything, and it messes with the formatting of the lyrics.

Yes that should have been implemented, I also made a function for it but somehow never called it? Thanks for pointing this out!

huhridge:
> > I tested this, and it works great! I was gonna add this, but this is better than I would have implemented it.
> > Is there a possibility of adding a check to only translate when the song is actually Japanese? Right now, it passes everything, and it messes with the formatting of the lyrics.
> 
> Yes that should have been implemented, I also made a function for it but somehow never called it? Thanks for pointing this out!

Just tried it, and works perfectly now! You did call the function once to check whether to display the translation menu iirc.

41pha1:
> prettify your code with prettier but with config from this repo 🙄

Looks like its all working now :D

ririxidev:
Wouldn't it be better to include https://github.com/hexenq/kuroshiro js files in our repo and then invoke stuff from Translator.js after loading them by jsdeliver?

41pha1:
> Wouldn't it be better to include https://github.com/hexenq/kuroshiro js files in our repo and then invoke stuff from Translator.js after loading them by jsdeliver?

Probably however I think it would be difficult to achieve, as this also uses the analyzer functions here https://github.com/hexenq/kuroshiro-analyzer-kuromoji
and I have rewritten a decent amount of code that somehow seemed to be broken in the original repo (I think due to JavaScript version changes, I am not quite sure. Even the demo page at https://kuroshiro.org/ seems to be broken now) 
as well as added support for loading the dictionaries dynamically.

ririxidev:
Could you provide source code for this then which I will upload to repo?

41pha1:
Do you mean the code of the modified kuroshiro / analyzer bundle that currently sits along the Translator class in Translator.js?
I sadly don't have separated files anywhere as I exported everything directly to the bundle with npm.

I am not quite sure if I understand correctly :/

ririxidev:
So, you don't have any source-code for this code? Why's that? We will not be able to update a code then. We can't extract source from webpack compiled code. If you deleted it from disk you should be able to restore it with DMDE

41pha1:
Oh yea my bad! I added the unminified code and split it into kuromoji.js and kuroshiro.js If wanted we can move them to the separate repo?

ririxidev:
Revert this commit please. I didn't mean unminified code but source code of it. This is still bundled by webpack which is not a source code

41pha1:
I am afraid I cant provide you with what you are looking for in that case

kyrie25:
> I am afraid I cant provide you with what you are looking for in that case

We are asking if you can provide for us the repository/package in which you forked those files. If those codes are not owned by you, legally we cannot use it as we may risk breaking the terms of the license in which the code is published under.

41pha1:
> We are asking for the original code in which this was compiled to, if that clarifies anything for you.
> 
> Currently what you have committed here is the minified/compiled code from a Webpack build process, of which we cannot inspect, verify and maintain its functionality.

I will have to get to work on the files from the Kuroshiro repo again, because I was as naive as to modify the compiled code from npm directly and thus am not and never haven been in the possession of the uncompiled code.

huhridge:
Is there a possibility of bundling the translate function within Spicetify? So maybe it can be used in other extensions and custom apps, etc?

kyrie25:
> Is there a possibility of bundling the translate function within Spicetify? So maybe it can be used in other extensions and custom apps, etc?

The extension will be released as a standalone package so feel free to use that instead.

huhridge:
Yeah, but the problem is importing modules in the extensions and making it work with marketplace, unless that was fixed?

kyrie25:
> Yeah, but the problem is importing modules in the extensions and making it work with marketplace, unless that was fixed?

That would be discussed on the Marketplace side of development. You can always choose to opt out on having users install it via Marketplace and do it locally instead. We are also planning to make an option for developers to redirect users to their repository's Readme instead of installing by default (spicetify/spicetify-marketplace#346)
Either way, we are not forcing users to increase their installation package size by ten-fold for a feature that has probably around 1-2% of the userbase using it.

Pithaya:
In case it helps, I've also been trying to integrate Kuroshiro in a spicetify extension without having to download the dictionary files locally, and got it to work without having to host the files in a repo. You can directly get the dictionary files from npm.
There is an [issue](https://github.com/takuyaa/kuromoji.js/issues/37) with loading external urls however (I guess this is why the library's code had to be edited here to use downloaded dictionary files directly?), so I had to add a fix (found [here](https://github.com/mobilusoss/textlint-browser-runner/pull/7/files)) to get it working.

The repo is private for now so I can't link the file but here is the code I'm using (the extension is built with spicetify creator):
```ts
import Kuroshiro from 'kuroshiro';
import KuromojiAnalyzer from '../../node_modules/kuroshiro-analyzer-kuromoji/dist/kuroshiro-analyzer-kuromoji.min.js';

type ExtendedXMLHttpRequest = XMLHttpRequest & {
    patched_open: (method: string, url: string) => void;
};

const dictPath = 'https://cdn.jsdelivr.net/npm/kuromoji@0.1.2/dict';
const dictPathWithoutDoubleSlash =
    'https:/cdn.jsdelivr.net/npm/kuromoji@0.1.2/dict';
/**
 * Fix an issue with kuromoji when loading dict from external urls
 * See: https://github.com/takuyaa/kuromoji.js/issues/37
 * Adapted from: https://github.com/mobilusoss/textlint-browser-runner/pull/7
 */
export function applyKuromojiFix(): void {
    (XMLHttpRequest.prototype as ExtendedXMLHttpRequest).patched_open =
        XMLHttpRequest.prototype.open;
    XMLHttpRequest.prototype.open = function (method: string, url: string) {
        if (url.indexOf(dictPathWithoutDoubleSlash) === 0) {
            (this as ExtendedXMLHttpRequest).patched_open(
                method,
                url.replace(dictPathWithoutDoubleSlash, dictPath)
            );
        } else {
            (this as ExtendedXMLHttpRequest).patched_open(method, url);
        }
    };
}

export async function createKuroshiro(): Promise<Kuroshiro> {
    applyKuromojiFix();

    const kuroshiro: Kuroshiro = new Kuroshiro();
    const analyzer: KuromojiAnalyzer = new KuromojiAnalyzer({
        dictPath: dictPath,
    });

    await kuroshiro.init(analyzer);

    return kuroshiro;
}

```

Overrevvv:
Thank you for your contribution and this fantastic extension. I'd like to know if we can convert Romanji further into English.

ririxidev:
@Pithaya if you want to integrate this with lyrics-plus, please create a pull request. This one is on hold because pr's author can't provide source code with his changes - and since your code can get dictionary files from jsdelivr, it's even better for us 👍

ohitstom:
i really hope this isnt abandoned lol

41pha1:
I think I got it to work by loading kuroshiro, kuromoji and the dictionary files via JSDeliver directly from npm using the XMLHttp fix similar to the one suggested by [Pithaya](https://github.com/Pithaya).

ririxidev:
It's okay by me, @41pha1

kyrie25 (approved):
LGTM

ririxidev (approved):

Tetrax-10:
@41pha1 

In line [114](https://github.com/spicetify/spicetify-cli/blob/master/CustomApps/lyrics-plus/OptionsMenu.js#L114) should that be `isJapanese` or `!isJapanese` Because if its a japanese song then the menu should show up right?

image

PR #1991, opened 2022-10-23
feat(github/actions): lint pr title

ririxidev:
This pull request includes:
 - Adding Github Actions to lint Pull Request title to match [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specs
 - Moving `CODEOWNERS` to `.github` folder where it should have been from beginning


kyrie25 (approved):

theRealPadster (approved):

CharlieS1103 (approved):

.github/workflows/lintpr.yml
```diff
@@ -0,0 +1,17 @@
...
+            - synchronize
+
+jobs:
+    lintpr:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Lint pull request title
+              uses: amannn/action-semantic-pull-request@v5
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
afonsojramos, line 17: I'd prefer to just have a regex instead so that we have more control over the action itself. For example. I don't believe we always need a scope, but if we do, we should define which are the possible scopes.
```suggestion
            - name: Validate PR title
               uses: actions/github-script@v6
               with:
                 script: |
                   const title = context.payload.pull_request.title;
                   const regex = /^(?<type>build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(?<scope>\(\w+\)?((?=:\s)|(?=!:\s)))?(?<breaking>!)?(?<subject>:\s.*)?/gm;
                   const match = regex.exec(title);

                   if (!match) {
                     core.setFailed('Invalid PR title, please follow conventional-commits' guidelines');
                   }
                   if (!match.groups.type && !match.groups.subject) {
                     core.setFailed('Missing type and subject in PR title');
                   }
                   if (!match.groups.type && match.groups.subject) {
                     core.setFailed('Missing type in PR title');
                   }
                   if (match.groups.type && !match.groups.subject) {
                     core.setFailed('Missing subject in PR title');
                   }
```

afonsojramos (changes requested):
I'd prefer that this was just a

.github/workflows/lintpr.yml
ririxidev: image
Sure, however this regex should only allow lower-case letters and I don't understand this part - `?(?<breaking>!)?`

afonsojramos (approved):

PR #1994, opened 2022-10-26
fix: add support for `1.1.97`

kyrie25:
- Compatibility with/hotfix for Spotify `1.1.97`
- Fixes issue where launching FAD with Lyrics Plus enabled while on home page crashes the client on the aforementioned version.

Backwards compatible & tested.

### DIFF
--- a/CustomApps/lyrics-plus/index.js
+++ b/CustomApps/lyrics-plus/index.js
@@ -534,28 +534,29 @@ class LyricsContainer extends react.Component {
                 react.createElement(AdjustmentsMenu, { mode })
             ),
             activeItem,
-            react.createElement(TopBarContent, {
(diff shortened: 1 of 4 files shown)


theRealPadster (approved):

ririxidev (approved):

ririxidev:
Backwards compatibility tested to which version?

kyrie25:
> Backwards compatibility tested to which version aka what is the last working version?

`1.1.90` and above, as officially stated on the CLI. Experimental Features requires `1.1.92` as mentioned in #1965.

ririxidev:
Tested `1.1.85` right now and it works it seems and such we still have full backwards compatibility with your fix for `1.1.97` resulting in `12` supported versions
image

PR #1999, opened 2022-10-30
docs: Updating README.md to be more precise

max12mx13:
Its just not very obvious in my opinion - happy for the text to be changed to whatever you see sensible!

### DIFF
--- a/CustomApps/lyrics-plus/README.md
+++ b/CustomApps/lyrics-plus/README.md
@@ -23,6 +23,7 @@ Lyrics in Unsynced and Genius modes can be search and jump to. Hit Ctrl + Shift 
 ![search](./search.png)
 
 Customise colors, change providers' priorities in config menu. Config menu locates in Profile Menu (top right button with your user name).
+> Note - You have to be in the lyrics section of the spotify app for these settings to show
 
 To install, run:
 


max12mx13:
Sorry about all the title changes

kyrie25:
Config only displays while you are on the page is self-explanatory.
Settings for an app only needs to be displayed while that app is opened and where the changes can be seen live.
This logic applies for all custom apps and extensions which has its own layout that Spicetify has by default (FAD, Reddit, Lyrics Plus).

So while we appreciate your contribution, this is an unnecessary commit that doesn't really serve any purpose.

CharlieS1103:
I don't actually think it's that clear to be entirely obvious, it seems a bit strange that it randomly appears in the profile menu with zero context or anything provided, so it doesn't seem like a bad idea to inform users of this. @kyrie25

kyrie25:
> it seems a bit strange that it randomly appears in the profile menu with zero context or anything provided

Those settings are only needed while the app is opened and thus only displayed when it is put into effect, that itself is not *zero context*


Pointing out that the settings modal for an app *is* available while using *that same app* is very much obvious. If you do want to inform the users that a config modal exists, then make a complete section about it so it does serve a purpose.

max12mx13:
Where would you suggest we add this section?
Would a README in the CustomApps dir work?

kyrie25:
> Where would you suggest we add this section?
> 
> Would a README in the CustomApps dir work?

This same README at the place the change was added would be fine.

max12mx13:
So what do you suggest to add. I'm not sure

CharlieS1103:
@kyrie25 it just seems odd to me that we would put that in the profile menu, either it should stay permanently inside the profile menu or we should have a second (more visible) settings button that only appears when lyrics plus is opened, such as we have on marketplace. Most users, including myself, aren't used to seeing the profile menu gain and lose options dependent on which page you are on.

kyrie25:
> it just seems odd to me that we would put that in the profile menu, either it should stay permanently inside the profile menu

Scope of where the settings is applied, already said above. There's no reason to change Spotify settings outside the Spotify application where it is not put into use, same reason for that.

> we should have a second (more visible) settings button that only appears when lyrics plus is opened, such as we have on marketplace. Most users, including myself, aren't used to seeing the profile menu gain and lose options dependent on which page you are on.

Lol skill issue. There has only been one user came in and asks where the config is, *most users* don't have the issue you claim they have, for as long as these default Spicetify custom apps been a thing.

max12mx13:
Ok, I just think its not very clear, no need to be rude.

kyrie25:
> Ok, I just think its not very clear, no need to be rude.

That wasn't addressed to you.

PR #2013, opened 2022-11-13
fix(lyrics-plus): make invisible lyrics unclickable

huhridge:


huhridge:
what happened here?

kyrie25 (approved):
LGTM

CustomApps/lyrics-plus/style.css
```diff
@@ -611,4 +612,4 @@ button.switch.small {
 }
 .lyrics-lyricsContainer-UnsyncedLyricsPage .split > div {
 	flex: 50%;
-}
+}
```
kyrie25, line 615: > what happened here?
@ririxidev may have the fix for the workflow stuff, for now you can return new line at EOF here and it should be good

kyrie25 (approved):

ririxidev (approved):

PR #2025, opened 2022-11-24
fix(custom-apps): support for `Your Library X`

kyrie25:
In preparation for the `Your Library X` internal feature. It has been under development for around 10 versions now and is most likely will soon be put into production, so we should probably make our own patches ahead.
Spotify_3Qm2m4wHKf
Spotify_JuBIH0RU11

To do:
- [x] Add compatibility for older Spotify layout
- [x] ~~Make custom apps listing look better aesthetic-wise~~ I guess in a way now?
- [ ] ~~Use Spotify's carousel (if possible)~~ Not worth the trouble.


kyrie25:
Yeah Spotify completely broke my patch on `1.2.0`, this was futile

PR #2034, opened 2022-11-30
changed N to n 😅

LiReNa00:


CharlieS1103:
Personally I think we should merge this, unless there's a reason we use capital N

LiReNa00:
now dont ask me why I am reopening the request again and gain I messed up real bad 😥

ririxidev:
`N` is capitalized because it's default answer in a code
image
and such I believe it shouldn't be changed

theRealPadster:
The capital indicates a default.

CharlieS1103:
Ah I see, never knew that lol @theRealPadster

LiReNa00:
oh then will Y will be good? I am not sure

LiReNa00:
XD I wonder

LiReNa00:
Oh that is the case .... ok

PR #2038, opened 2022-12-04
fix(apply): define custom app routes

kyrie25:
Fixes issues where custom apps that uses routes (like Marketplace Readme) was not working properly


CharlieS1103:
@kyrie25 Have you tested this with custom apps that don't use any routes or does it not matter?

kyrie25:
> @kyrie25 Have you tested this with custom apps that don't use any routes or does it not matter?

Default ones works just fine.

theRealPadster (approved):

CharlieS1103 (approved):

PR #2058, opened 2022-12-18
fix(github/actions): AUR update

itsmeow:
For some reason this refuses to redirect/proxy from the old URL on my side so I'm just going to change it here. I made some server changes that broke some of my webserver stuff, sorry about that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,7 +92,7 @@ jobs:
     name: Trigger Homebrew/AUR Release
     needs: release
     steps:
-      - name: Update Homebrew/AUR
+      - name: Update AUR package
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -95,5 +95,5 @@ jobs:
       - name: Update Homebrew/AUR
```
theRealPadster, line 95:
```suggestion
      - name: Update AUR package
```
Since we're going to do the homebrew one with a separate action.

theRealPadster (changes requested):

ririxidev (approved):

theRealPadster (approved):

PR #2075, opened 2022-12-29
fix(lyrics-plus): copy converted Japanese lyrics

kyrie25:
lol
image


ririxidev (approved):

afonsojramos (approved):

afonsojramos (changes requested):
I think we could extract this code to a function to remove its duplication, no?

kyrie25:
> I think we could extract this code to a function to remove its duplication, no?

Yeah I suppose so, will look into it later.

afonsojramos (approved):

PR #2082, opened 2023-01-03
fix(spicetifyWrapper):  use optional chaining

CharlieS1103:


CharlieS1103:
Note, this still doesn't fix the issue with libraryX, there's a react minified error as well.

kyrie25:
This is supposed to throw an error to notify that there is a very crucial problem with the regex or mappings lol, also React would still throw an error on null element regardless.

kyrie25:
> Note, this still doesn't fix the issue with libraryX, there's a react minified error as well. 

This still works on `1.2.2` for me

PR #2085, opened 2023-01-06
chore: ignore spicetify binaries

vegerot:
(gitignore) ignore spicetify-cli executable

Summary: Ignore the `spicetify-cli` executable generated from `go build` on Unix

Test Plan: N/A

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/spicetify/spicetify-cli/pull/2085).
* __->__ #2085

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # Executables
 bin/
+spicetify-cli
+spicetify
 *.exe
 
 # MacOS


CharlieS1103:
> I don't exactly see the point in this PR. This is not, and never has been, a prevalent issue.
> If you don't want to commit the executable you can choose not to stage it.

I mean, that might be true but isn't the purpose of gitignore literally exactly for stuff like this, that no one will ever want to commit so you may as well make it easier for them to do so. Your point is that it's easy to not commit, but that doesn't mean this PR shouldn't be accepted just to make it clear to newer devs that they shouldn't commit it and such.

kyrie25:
> Your point is that it's easy to not commit, but that doesn't mean this PR shouldn't be accepted just to make it clear to newer devs that they shouldn't commit it and such.

Again, this has never been an issue, and we have no shortage of new contributors over the months to prove that.
If you're really wanting to stick it to the new devs then at least make it follow the guides you wrote for them.

.gitignore
```diff
@@ -1,5 +1,6 @@
 # Executables
 bin/
+spicetify-cli
```
kyrie25, line 3:
```suggestion
spicetify
```
image

kyrie25 (dismissed):

CharlieS1103:
What do you mean?? All I'm saying is that your argument is that it's never been an issue but it doesn't mean that making things more convenient and clear, albeit slightly, is a bad thing. There's simply no good reason not to accept this pull request, and a few solid ones to.

.gitignore
ririxidev: add both instances then...

```suggestion
spicetify-cli
spicetify
```

CharlieS1103: Maybe we can have it ignore both, as either one of them could be what's generated depending on whether someone read through the instructions or not

CharlieS1103: Ririxi got to it first lol

kyrie25:
> What do you mean?? All I'm saying is that your argument is that it's never been an issue but it doesn't mean that making things more convenient and clear, albeit slightly, is a bad thing.

Can you point me to exactly I said this was a bad change?
My argument was that it is a pointless change directing at a non-issue.

CharlieS1103 (approved):
Looks all good to me

vegerot:
@CharlieS1103 I don't have permission to merge this

ririxidev:
Resolve conflicts

vegerot:
> Resolve conflicts

done!

PR #2086, opened 2023-01-06
docs(devtools): add macOS instructions

vegerot:
(docs) Add macOS instructions for enable-devtools

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -285,7 +285,7 @@ restore             Restore Spotify to original state.
 clear               Clear current backup files.
 
 enable-devtools     Enable Spotify's developer tools.
-                    Hit Ctrl + Shift + I in the client to start using.
+                    Hit Ctrl + Shift + I (Windows) or Cmd + Option + I (macOS) in the client to start using.
 
 watch               Enter watch mode.
                     On default, update CSS on color.ini or user.css's changes.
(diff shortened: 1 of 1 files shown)


CharlieS1103 (approved):
Seems good to me

theRealPadster (changes requested):
If you're specifying each OS, then you need to mention Linux as well. I'd recommend just saying "ctrl/cmd + shift + I"

vegerot:
> If you're specifying each OS, then you need to mention Linux as well. I'd recommend just saying "ctrl/cmd + shift + I"

but it's not shift, it's option (alt)

`ctrl/cmd + shift/option + I` sounds weird xD

theRealPadster:
> but it's not shift, it's option (alt)
> 
> `ctrl/cmd + shift/option + I` sounds weird xD

Ahh that's why I could never get the shortcut to work on macos 😂 then I'd just add a "/Linux" to your "(Windows)"

spicetify.go
```diff
@@ -285,7 +285,7 @@ restore             Restore Spotify to original state.
 clear               Clear current backup files.
 
 enable-devtools     Enable Spotify's developer tools.
-                    Hit Ctrl + Shift + I in the client to start using.
+                    Hit Ctrl + Shift + I (Windows) or Cmd + Option + I (macOS) in the client to start using.
```
theRealPadster, line 288:
```suggestion
                    Press Ctrl + Shift + I (Windows/Linux) or Cmd + Option + I (macOS) in the Spotify client to start using.
```

theRealPadster (changes requested):
I think this addresses all the feedback.

vegerot:
@theRealPadster I don't have permission to merge this PR

PR #2092, opened 2023-01-11
feat(installer): write log to file for support

FivePixels:
A user in the Discord server was having issues with the installer not automatically setting up their shell's PATH to the Spicetify command line executable. I proposed to @theRealPadster that we should start logging installs so if this does happen again, we can review the logs and determine where the problem resides.

This feature:
- Implements a log function in the installer script that continues to log to the standard output as normal, but also logs to a `install.log` file. Each log in the `install.log` file includes a timestamp for convenience.
- Resets the `install.log` file each time the installer script is ran. In hindsight, this could probably be removed if this script gets ran each time Spicetify is updated (via AUR, or whatever package manager/method). We can retain a log of this on machine over time so we can tell when an issue started occurring.  
- Adds the `install.log` file to `.gitignore` to prevent it from being picked up in commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ bin/
 node_modules
 package-lock.json
 package.json
+
+# Logs
+install.log
--- a/install.ps1
+++ b/install.ps1
@@ -10,9 +10,34 @@ if ($v) {
     $version = $v
 }
 
+$log_file_dir = "$sp_dir\install.log"
+
+# Function to log output to file.
+Function Log {
+  Param([String] $ActionText, [String[]] $Texts, [Boolean[]] $Emphasized, [Boolean] $NewLine)
+  if (-not (Test-Path $log_file_dir)) {
+    New-Item -Path $log_file_dir -ItemType File -Force | Out-Null
+  }
+  if (-not ([String]::IsNullOrEmpty($ActionText))) {
+    $FormattedActionText = "{0, -15}" -f $ActionText
+    Write-Part $FormattedActionText
+  }
+  $LogText = $FormattedActionText
+  for ($i=0; $i -lt $Texts.Length -AND $Texts.Length -eq $Emphasized.Length; $i++) {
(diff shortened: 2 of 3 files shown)


theRealPadster (approved):
LGTM 👍

kyrie25 (approved):
LGTM! Could you possibly do this with the Windows install script also since it's good to have diagnostics information nonetheless.

FivePixels:
> LGTM! Could you possibly do this with the Windows install script also since it's good to have diagnostics information nonetheless.

I've been digging into the Powershell script and it's going to need to be rewritten. How we currently log to the standard output builds out a log message part-by-part. My refactor plan is to associate an 'action' with each log line (since that's what happens when you run the script) and create the log message using a Log function similar to the one in `install.sh`.

FivePixels:
My original plan was to implement a function that accepted a function as a parameter, so all we would be doing is updating all of the standard output writes to `Log $Function:WritePart "12345"` and so on, but the issue with that is it spawns a new line for each Log call in the log file. This isn't reflective of what actually happens in standard output.

FivePixels:
The best approach I could get was passing in two arrays that we do an index loop over. The two arrays have matching indices for texts, and whether that text should be emphasized or not. Since we were logging before actually creating our Spicetify directory that contains the install.log, I opted to make the Log function create the file and folder if it doesn't exist. This allows us to log the folder creation in both the standard output and the log file.

I really don't like PowerShell after this :)

ririxidev (approved):

FivePixels:
@theRealPadster what about the curl Powershell script? Did it also need updated?

theRealPadster:
> @theRealPadster what about the curl Powershell script? Did it also need updated?

I forgot it existed. I'm not sure why there are two tbh. The other day on discord someone was trying to use it and it wasn't working so it was recommended that they use the regular one.

PR #2122, opened 2023-01-24
fix(lyrics-plus): inject Translator lib when translate option is enabled

Tetrax-10:
Changes:
1. Injects the Translator lib when translate option is enabled
2. injects the lib only once
3. fixed menu condition from `!isJapanese` to `isJapanese`

# Bug specified in `2`
image

### DIFF
--- a/CustomApps/lyrics-plus/OptionsMenu.js
+++ b/CustomApps/lyrics-plus/OptionsMenu.js
@@ -88,6 +88,8 @@ const OptionsMenu = react.memo(({ options, onSelect, selected, defaultValue, bol
 const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, isJapanese, hasNeteaseTranslation }) => {
 	if (!showTranslationButton) return null;
 
+	let translator = new Translator();
+
 	let menuOptions = null;
 	if (isJapanese) {
 		menuOptions = {
@@ -111,7 +113,7 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 				Spicetify.ReactComponent.Menu,
 				{},
 				react.createElement("h3", null, " Conversions"),
-				translatorLoaded || !isJapanese
+				translatorLoaded || isJapanese
 					? react.createElement(OptionList, {
 							items: [
 								{
@@ -134,6 +136,8 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 								CONFIG.visual[name] = value;
 								localStorage.setItem(`${APP_NAME}:visual:${name}`, value);
 								lyricContainerUpdate && lyricContainerUpdate();
+								CONFIG.visual[name] ? Spicetify.showNotification("Translating...", false, 500) : null;
+								translator.includeExternal("inject");
 							}
 					  })
 					: react.createElement(
--- a/CustomApps/lyrics-plus/OptionsMenu.js
+++ b/CustomApps/lyrics-plus/OptionsMenu.js
@@ -113,39 +113,32 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 				Spicetify.ReactComponent.Menu,
 				{},
 				react.createElement("h3", null, " Conversions"),
-				translatorLoaded || isJapanese
-					? react.createElement(OptionList, {
-							items: [
-								{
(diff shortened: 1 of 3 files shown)


CustomApps/lyrics-plus/OptionsMenu.js
```diff
@@ -111,7 +113,7 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 				Spicetify.ReactComponent.Menu,
 				{},
 				react.createElement("h3", null, " Conversions"),
-				translatorLoaded || !isJapanese
+				translatorLoaded || isJapanese
```
lixiao189, line 116: I think that using `translatorLoaded || !isJapanese` is better. Because I add netease music translation in this custom app. When spotify play an English songs, the translator won't loaded the Japanese translator, however it may load Chinese translation from netease music. So it should loaded the menu option immediately.

CustomApps/lyrics-plus/OptionsMenu.js
```diff
@@ -134,6 +136,8 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 								CONFIG.visual[name] = value;
 								localStorage.setItem(`${APP_NAME}:visual:${name}`, value);
 								lyricContainerUpdate && lyricContainerUpdate();
+								CONFIG.visual[name] ? Spicetify.showNotification("Translating...", false, 500) : null;
+								translator.includeExternal("inject");
```
rxri, line 140: Could it be in a new function instead?
```javascript
translator.injectExternals();
```

rxri (changes requested):

CustomApps/lyrics-plus/OptionsMenu.js
Tetrax-10: you mean this part as a separate fuction? then ofcourse. Do you want me to create a separate fuction to trigger that inject method instead of that if else thing?
image

rxri: That what would be the best I guess?
```javascript
injectExternals() {
	this.includeExternal(kuroshiroPath);
	this.includeExternal(kuromojiPath);
}
```

Tetrax-10:
Done

CustomApps/lyrics-plus/OptionsMenu.js
```diff
@@ -111,37 +113,32 @@ const TranslationMenu = react.memo(({ showTranslationButton, translatorLoaded, i
 				Spicetify.ReactComponent.Menu,
 				{},
 				react.createElement("h3", null, " Conversions"),
-				translatorLoaded || !isJapanese
```
rxri, line 114: why this condition has been removed?

rxri (changes requested):

CustomApps/lyrics-plus/OptionsMenu.js
Tetrax-10: That condition is there to toogle this popup
picture 1:
image

That `isJapanese` condition is always useless as we can't open that popup if the song is not japanese because this button will be hiden if the song is not japanese. Even in current lyrics-plus `isJapanese` is useless so forget `isJapanese` variable.
image

Now `translatorLoaded`
If the main functions of external lib is not available `translatorLoaded` will become `false` and loading screen like this will popup instead of the menu (Picture 1). If `translatorLoaded` is `true` (main functions of external lib is available) menu will appear like picture 1.
image

I removed `translatorLoaded` condition as lyrics plus wont inject the external lib as default. Instead it will check if convert option is enabled, if enabled then it will inject external lib at startup. so if convert option is diabled external lib will not be injected.
image

If a user enables convert by clicking that convert option. `onclick`(onChange) the external lib are injected and the lyrics are translated
Untitled


Lets say if the user disables convert and closes spotify and relaunch and then opens the menu to enable it again.
If i didnt remove `translatorLoaded` condition user can never enable the covert option again as user will only see a popup that says loading as `translatorLoaded` will be `false` as the external lib will not be available as convert option is disabled

rxri: that makes sense, my bad

kyrie25 (approved):

Tetrax-10:
Why isn't this merged automatically?

PR #2127, opened 2023-01-28
fix(installer): incorrect curl usage

CreepTron365:
The current installer script for MacOS and Linux does not use curl correctly to find the latest version url. This script has fixed this issue.

Edit: This issue is also seen with the marketplace script.


rxri:
That doesn't seem right. Old method works perfectly fine like it should be - your change breaks getting tag from JSON response
image
If response is HTML response, it's something on GitHub end and not an issue in script

PR #2128, opened 2023-01-28
feat(extensions): Added Extension to play songs with ENTER

nicolapace:
Simple Extension that plays the first song in the search list or goes inside the first podcast by just pressing the ENTER button on the keyboard.


kyrie25:
If you wish to publish an extension, please follow this guide for publishing on Marketplace:
https://github.com/spicetify/spicetify-marketplace/wiki/Publishing-to-Marketplace

We do not accept new extensions/custom apps to be added to the CLI as these are meant to be legacy and bundled with every installation.

PR #2129, opened 2023-01-29
fix: patch for `1.2.4`

kyrie25:
Compatibility patch for Spotify `1.2.4` (im tired)
Listed changes/breaks & fixes:
- Spotify removed the `URI` method and broke it down to several small components/hooks. This patch aims to inject frequently used methods of the `URI` class and creates a backwards-compatible method for extensions/custom apps.
image

- Added `ReactComponent` to the list of properties to test using `Spicetify.test` for ease of debugging
- Fixed hooks for sub menu components
image

Works on `1.1.90` and higher, have not tested below.

### DIFF
--- a/CustomApps/new-releases/Card.js
+++ b/CustomApps/new-releases/Card.js
@@ -2,8 +2,8 @@ class Card extends react.Component {
 	constructor(props) {
 		super(props);
 		Object.assign(this, props);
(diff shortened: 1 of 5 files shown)


afonsojramos (approved):

rxri (approved):
`1.1.85` works as well

PR #2175, opened 2023-02-27
fix(workflows/winget): use new token

afonsojramos:
This token should only be used for this as it only has the `public_repo` scope.


rxri (approved):

kyrie25 (approved):

sitiom:
Now it works:
- https://github.com/microsoft/winget-pkgs/pull/101118

Issue #2174, opened 2023-02-26
Possible to add translation of Chinese characters


comcomson:
### Chinese characters conversation in the Lyrics Plus APP

Thank you Spicetify & Lyrics Plus developers for making such great extension!

For the Lyrics Plus App,

In Chinese, we have traditional/simplified characters. When a Hong Kong Cantonese song is being played, sometimes the Lyric Plus extension will fetch a Simplified Version of Chinese lyrics. It may be possible to add a translation button to automatically convert Simplified Characters (used in Mainland China) to Traditional Characters (Used in Hong Kong/ Taiwan).

There is already a "[opencc-js](https://github.com/nk2028/opencc-js)" on GitHub which easily convert Chinese characters w/ a JS.
For Traditional to Simplified Characters: https://cdn.jsdelivr.net/npm/opencc-js@1.0.5/dist/umd/t2cn.js
For Simplified to Traditional: https://cdn.jsdelivr.net/npm/opencc-js@1.0.5/dist/umd/cn2t.js
Full conversation JS: https://cdn.jsdelivr.net/npm/opencc-js@1.0.5/dist/umd/full.js

I believe this feature will greatly increase the functionality of the Lyrics Plus app.

PR #2184, opened 2023-03-04
fix(lyrics-plus): differentiate zh and jp

Deskehs:
The current Japanese detector doesn't differentiate Chinese and Japanese for some reason and bugs me a lot, I tried to fix it by using an external language detector module franc. In this current form works but due to me botching I still needs some time to fix the remaining issues.

This, in theory, could enable #2174 by adding in OpenCC and some changes (which I will not do in this PR).

Before:
Spotify_pURYOhO0aP
Spotify_s6vMhJBO1Z

After:
Spotify_riKH4Fw0fH
Spotify_IkWoi4MxXJ


Deskehs:
Checked some other alternatives on how to solve this problem, maybe I could ditch franc and search if all the lyrics uses kana frequently

Deskehs:
Ok, using the alternative solution right now, I'm weighting that Chinese would be more than 60% Hanzi and less than 40% Kana (if there is a chance that a song is referencing something in Japanese), which also solves the other bug I'm also dealing about to.

Deskehs:
I'm starting to add a threshold to this, default is 0.2 as 0.6-0.4=0.2, below 0 is most likely Japanese, above the threshold is most likely Chinese, the range is from -1 to 1.

I'll add a setting later to change the threshold from 0-1.

Deskehs:
Added an option
Spotify_sGo57lMZf1

The only thing left to do is to solve the floating point issue.

Deskehs:
It's ready to review after some testing.

CustomApps/lyrics-plus/index.js
```diff
@@ -97,6 +98,8 @@ CONFIG.visual["lines-before"] = parseInt(CONFIG.visual["lines-before"]);
 CONFIG.visual["lines-after"] = parseInt(CONFIG.visual["lines-after"]);
 CONFIG.visual["font-size"] = parseInt(CONFIG.visual["font-size"]);
 
+CONFIG.visual["detect-threshold"] = parseInt(CONFIG.visual["detect-threshold"])
```
kyrie25, line 101:
```suggestion
CONFIG.visual["detect-threshold"] = parseInt(CONFIG.visual["detect-threshold"])
```

CustomApps/lyrics-plus/Utils.js
```diff
@@ -56,10 +56,37 @@ class Utils {
 	}
 
 	static isJapanese(lyrics) {
-		for (let lyric of lyrics)
-			if (/[\u3000-\u303F]|[\u3040-\u309F]|[\u30A0-\u30FF]|[\uFF00-\uFFEF]|[\u4E00-\u9FAF]|[\u2605-\u2606]|[\u2190-\u2195]|\u203B/g.test(lyric.text))
-				return true;
-		return false;
+		let rawLyrics = "";
+
+		for (let lyric of lyrics) rawLyrics = rawLyrics + ` ${lyric.text}`;
```
kyrie25, line 61:
```suggestion
		const rawLyrics = lyrics.map(line => line.text).join(" ");
```

CustomApps/lyrics-plus/Settings.js
```diff
@@ -475,7 +475,16 @@ function openConfig() {
 					key: "highlight-color",
 					type: ConfigInput,
 					when: () => !CONFIG.visual["colorful"]
-				}
+				},
+				{
+					desc: "Text convertion: Detection threshold",
+					info: "(Distinguish between Japanese and Chinese; in percentage)",
```
kyrie25, line 481: Please be more descriptive of the settings option so that new users could understand
Threshold/Percentage of what?

CustomApps/lyrics-plus/Utils.js
```diff
@@ -56,10 +56,37 @@ class Utils {
...
+			}
+		} else {
+			return false;
+		}
+
+		if (((cjkCount / charMatch.length - kanaCount / charMatch.length + 1 ) / 2 * 100 ) >= CONFIG.visual["detect-threshold"]) {
+			return false;
+		} else {
+			return true;
+		}
```
kyrie25, line 89: Please refactor this part by splitting conditions into variables and simplify it like so.
```suggestion
		return !((cjkCount / charMatch.length - kanaCount / charMatch.length + 1 ) / 2 * 100 ) >= CONFIG.visual["detect-threshold"]
```

CustomApps/lyrics-plus/Utils.js
```diff
@@ -56,10 +56,37 @@ class Utils {
...
+
+		const kanaRegex = /[\u3001-\u3003\u3005\u3007\u301d-\u301f\u3021-\u3035\u3038-\u303a\u3040-\u30ff\uff66-\uff9f]/gu;
+
+		const cjkRegex = /\p{Unified_Ideograph}/gu;
+
+		const charMatch = rawLyrics.match(new RegExp(kanaRegex.source + "|" + cjkRegex.source, "gu"));
+
+		let cjkCount = 0;
+
+		let kanaCount = 0;
```
kyrie25, line 71:
```suggestion
		const kanaRegex = /[\u3001-\u3003\u3005\u3007\u301d-\u301f\u3021-\u3035\u3038-\u303a\u3040-\u30ff\uff66-\uff9f]/gu;
		const cjkRegex = /\p{Unified_Ideograph}/gu;
		const charMatch = rawLyrics.match(new RegExp(kanaRegex.source + "|" + cjkRegex.source, "gu"));

		let cjkCount = 0;
		let kanaCount = 0;
```

kyrie25 (changes requested):
Prettify changes.

CustomApps/lyrics-plus/Settings.js
Deskehs: I have changed the description as an attempt.

Deskehs:
The requested changes broke the detection, don't merge yet

Deskehs:
Should be fixed now

CustomApps/lyrics-plus/Settings.js
Deskehs: Just updated the description again.

kyrie25 (approved):
LGTM.

CustomApps/lyrics-plus/Utils.js
```diff
@@ -56,10 +56,28 @@ class Utils {
...
+			for (const character of charMatch) {
+				if (character.match(cjkRegex)) {
+					cjkCount++;
+				} else {
+					kanaCount++;
+				}
+			}
+		} else {
+			return false;
+		}
```
kyrie25, line 78: Seems like this can be further shortened also
```suggestion
		if (!charMatch) return false;
		
		for (const character of charMatch) {
			if (character.match(cjkRegex)) {
				cjkCount++;
			} else {
				kanaCount++;
			}
		}
```

Deskehs:
Don't mind the commit changes, just going for the style of commits this repo

Deskehs:
seems i messed up git, i'll make a new pull request for that

PR #2185, opened 2023-03-05
fix(lyrics-plus): improve detection of zh and ja

Deskehs:
A new pr continuing from #2184


kyrie25 (approved):

rxri (changes requested):
prettify all files

jsHelper/expFeatures.js
```diff
@@ -1,59 +1,59 @@
 (function () {
-    let overrideList,
```
kyrie25, line 2: Prettify and commit **only** the relevant files.

kyrie25 (changes requested):

Deskehs:
reverted the file to original status as requested

kyrie25 (approved):

CustomApps/lyrics-plus/Utils.js
```diff
@@ -56,10 +56,26 @@ class Utils {
...
+
+		for (const character of charMatch) {
+			if (character.match(cjkRegex)) {
+				cjkCount++;
+			} else {
+				kanaCount++;
+			}
+		}
+
+		return ((kanaCount / charMatch.length - cjkCount / charMatch.length + 1) / 2) * 100 >= CONFIG.visual["ja-detect-threshold"];
```
rxri, line 78: can't we split this line to variables so it's easier to read?

Deskehs: just separated the percentages into it's own variables

rxri:
Don't force push every-time you change something. It makes it harder to us to see what you're actually changing

Deskehs:
Oh oh alright then, is there anything you would like to change?

rxri (approved):

rxri (approved):

PR #2190, opened 2023-03-07
fix(home-config): patch for v2

kyrie25:
~~This thing has been broken for so long people forgot it existed lol~~
Works on `1.1.90` and higher, broken on `1.1.85` (as in the regex matched but they weren't using `home-v2`), have not tested in between.

`useEffect` implementation for menu item was removed because in some specific versions the `useEffect` hook was not present in the file.
Spotify_wUfFkolFF6

### DIFF
--- a/jsHelper/homeConfig.js
+++ b/jsHelper/homeConfig.js
@@ -9,8 +9,10 @@ SpicetifyHomeConfig = {};
 	let list;
 	// Store sections' statuses
 	const statusDic = {};
+	let mounted = false;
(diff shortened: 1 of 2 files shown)


rxri (approved):

cbyrneee:
@kyrie25 apologies for the off-topic comment, how did you get your spotify to look like that? Is it using experimental features, or some sort of plugin?

afonsojramos (approved):

PR #2198, opened 2023-03-15
feat(wrapper): use icons in menu item

Tetrax-10:
image

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -365,8 +365,9 @@ declare namespace Spicetify {
         /**
          * Create a single toggle.
          */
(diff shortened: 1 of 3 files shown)


globals.d.ts
```diff
@@ -365,8 +365,9 @@ declare namespace Spicetify {
...
         class Item {
-            constructor(name: string, isEnabled: boolean, onClick: (self: Item) => void);
+            constructor(name: string, isEnabled: boolean, onClick: (self: Item) => void, icon: Icon | string);
```
kyrie25, line 370: Passing in an icon should not be mandatory, nullish coalescing is already present in the wrapper.
```suggestion
            constructor(name: string, isEnabled: boolean, onClick: (self: Item) => void, icon?: Icon | string);
```

kyrie25 (changes requested):

globals.d.ts
Tetrax-10: sry I dont know much about `*.d.ts` files

theRealPadster (approved):
👍

PR #2205, opened 2023-03-18
fix(cmd/auto): not starting spotify

### LINKED ISSUE #2204: `spicetify auto` is not useful in its current state

- [X] I have checked the FAQ

```Markdown
- Spotify version: 1.2.7.1277.g2b3ce637
- Spicetify version: v2.16.2
```

### 📝 Description

The FAQ mentions:

> ### After Spotify's update, running `spicetify apply` or `spicetify update` breaks Spotify.
> After any Spotify update, always run `spicetify backup apply`.
> Optionally, make Spotify shortcut to run `spicetify auto` (instead of direct path to spotify executable), so that Spicetify can backup and apply, when it needs to, then launch Spotify automatically.
> 
> It may be the case that Spicetify is not supporting this new Spotify update yet. In that case, check our issue tracker.

However, this command does not launch Spotify as expected.
It will update and _restart_, if Spotify is already open. This doesn't make much sense given the suggestion that `spicetify auto` is a silver bullet for the _"how do I set and forget Spicetify so I don't have to interact with a terminal"_ problem.

I did a little digging and saw a fix for an unrelated issue made last year as the cause of this command becoming much less useful. It should be possible to do both; _not_ launch Spotify when calling `spicetify backup apply`, but _do_ launch it when calling `spicetify auto`.

### 📸 Screenshots

_No response_

SteffanDonal:


rxri (changes requested):
Please check my comment down below

spicetify.go
```diff
@@ -249,7 +249,7 @@ func main() {
 
 		case "auto":
 			cmd.Auto(version)
-			restartSpotify()
+			restartSpotify(true)
```
rxri, line 252:
```suggestion
			cmd.EvalSpotifyRestart(true)
```
would be much simpler fix

SteffanDonal: My reasoning for this is to avoid creating a special case where `noRestart` doesn't work as expected.

rxri: Well, auto is clearly to backup, apply and restart spotify so I don't believe `noRestart` should be checked there

kyrie25 (approved):

rxri (approved):

PR #2217, opened 2023-03-26
perf(lyrics-plus): disable conversion by default

Lseoksee:
Lyrics have been severely slow to load since the Japanese & Chinese text conversion option patch. Of course, if you disable the conversion option, it loads quickly, but I think you'd better disable the default value of the conversion option for those who don't listen to that music.

### DIFF
--- a/CustomApps/lyrics-plus/index.js
+++ b/CustomApps/lyrics-plus/index.js
@@ -48,7 +48,7 @@ const CONFIG = {
 		["font-size"]: localStorage.getItem("lyrics-plus:visual:font-size") || "32",
 		["translation-mode:japanese"]: localStorage.getItem("lyrics-plus:visual:translation-mode:japanese") || "furigana",
 		["translation-mode:chinese"]: localStorage.getItem("lyrics-plus:visual:translation-mode:chinese") || "cn",
-		["translate"]: getConfig("lyrics-plus:visual:translate"),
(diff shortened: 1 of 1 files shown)


CustomApps/lyrics-plus/index.js
```diff
@@ -48,7 +48,7 @@ const CONFIG = {
 		["font-size"]: localStorage.getItem("lyrics-plus:visual:font-size") || "32",
 		["translation-mode:japanese"]: localStorage.getItem("lyrics-plus:visual:translation-mode:japanese") || "furigana",
 		["translation-mode:chinese"]: localStorage.getItem("lyrics-plus:visual:translation-mode:chinese") || "cn",
-		["translate"]: getConfig("lyrics-plus:visual:translate"),
+		["translate"]: localStorage.getItem("lyrics-plus:visual:translate") || false,
```
kyrie25, line 51:
```suggestion
		["translate"]: getConfig("lyrics-plus:visual:translate", false),
```
image
`getConfig` returns a boolean, `localStorage.getItem` returns a string, which evaluates to `true` on `"false"`

kyrie25 (changes requested):

Lseoksee (commented):
that's a good correction.

kyrie25 (approved):
LGTM

rxri (approved):

PR #2218, opened 2023-03-26
fix(lyrics-plus): not using loops for detect algo

Deskehs:
This improves the detect language algorithm by not using loops. Maybe better performance? Idk, felt like it's better than looping every character.


kyrie25:
https://stackoverflow.com/questions/31459395/why-using-for-is-faster-than-some-or-filter


Filter does the same thing as loops, and is actually detrimental to performance (since it has to execute a function on top of the match), but with a margin so small it shouldn't even matter either way.

On those grounds, I don't believe this is a necessary change to commit, so while we do appreciate the effort I'll have to close this PR. 

Have a nice day!

Issue #1827, opened 2022-07-20
[Popup Lyrics] Window turns red on hover


LaeDev:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: For Windows Spotify 1.1.89.862.g94554d24
- Spicetify version: 2.10.2
```

### 📝 Description

popupLyrics screen goes red if i hover the cursor

### 📸 Screenshots

![image](https://user-images.githubusercontent.com/78576284/180052725-28e7648e-54e5-4656-a088-cffd4e1367f2.png)

kyrie25:
This is pretty much default behavior I think. You are requesting for it to be removed?

kyrie25:
As much as it does hurt my eyes when I hover over Popup Lyrics, this behavior is managed by the Spotify client and not the extension AFAIK, so I'm afraid we can't remove it.

abdansyakuuro:
19wwnefjfirnalalsh

kyrie25:
> This was fixed half a year ago in [electron](https://github.com/electron/electron/issues/34406), but only recently in [cef](https://github.com/chromiumembedded/cef/issues/3482) Waiting for the update from spotify devs, i guess!

Guess that solves the issue, we'll wait for a future release then!

PR #2221, opened 2023-03-27
feat: lyrics app improvement

Lseoksee:
first:
Modified the default musixmatch tokens for lyrics-plus and popupLyrics. The "popupLyrics" token `2005218b74f939209bda92cb633c7380612e14cb7fe92dcd6a780f` works, but the token lyrics-plus token `21051986b9886beabe1ce01c3ce94c96319411f8f2c122676365e3` does not seem to work properly.
This reduced the frequency of captcha error.

We applied two tokens, the existing token `2005218b74f939209bda92cb633c7380612e14cb7fe92dcd6a780f `and the newly issued token `22111116e5392ce08db76d4ce398786a1f2d7ba5c12a6184f28d46`

second:
Modified the default background color code for lyrics-plus to `8747370` If you turn on the colorful background option in "yrics-plus and look at the lyrics of the local music, it's too dark to see. If you have a better color, you can modify it.

before
Desktop Screenshot 2023 03 27 - 15 13 26 33

after
Desktop Screenshot 2023 03 27 - 15 11 15 88

### DIFF
--- a/Extensions/popupLyrics.js
+++ b/Extensions/popupLyrics.js
@@ -100,6 +100,8 @@ function PopupLyrics() {
 			const baseURL = `https://apic-desktop.musixmatch.com/ws/1.1/macro.subtitles.get?format=json&namespace=lyrics_synched&subtitle_format=mxm&app_id=web-desktop-app-v1.0&`;
 
 			const durr = info.duration / 1000;
+			const tokens = userConfigs.services.musixmatch.token.split("|");
+			const usertoken = tokens[Math.floor(Math.random() * tokens.length)];
 
 			const params = {
 				q_album: info.album,
@@ -109,7 +111,7 @@ function PopupLyrics() {
 				track_spotify_id: info.uri,
 				q_duration: durr,
 				f_subtitle_length: Math.floor(durr),
-				usertoken: userConfigs.services.musixmatch.token
+				usertoken
 			};
 
 			const finalURL =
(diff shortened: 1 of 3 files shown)


kyrie25:
> first:
> 
> Modified the default musixmatch tokens for lyrics-plus and popupLyrics. The "popupLyrics" token `2005218b74f939209bda92cb633c7380612e14cb7fe92dcd6a780f` works, but the token lyrics-plus token `21051986b9886beabe1ce01c3ce94c96319411f8f2c122676365e3` does not seem to work properly.
> 
> 
> 
> We applied two tokens, the existing token `2005218b74f939209bda92cb633c7380612e14cb7fe92dcd6a780f `and the newly issued token `22111116e5392ce08db76d4ce398786a1f2d7ba5c12a6184f28d46`
> 
> This reduced the frequency of captcha error.

This happens *partially* because the number of users using the same token 
hits rate limit. If we were to use the same one as `popupLyrics` then both extensions would face captcha.

Having another default API key for  a couple ten thousands of users would not change anything either.

We already linked a guide on the Spicetify docs and the Lyrics Plus app on how to retrieve their own API key so I believe this change should not be made.

kyrie25 (changes requested):
Previous comment

kyrie25:
> @kyrie25 Thank you for the good point. However, the existing Lyrics-plus token seems to have a problem. popupLyrics has no captcha error no matter how many songs you listen to, while Lyrics-plus has a captcha error with a probability of almost 90%.
> 
> I don't think it's just a matter of the number of users.

Popup Lyrics has a smaller user base than Lyrics Plus partially because it has a prominent issue that significantly affects UX (#1827)

Regardless, it is only one point on the comment I made and with a large influx of users on the scale of thousands using one default key, this would not make any difference.

kyrie25:
The changes on fallback color is much appreciated though, so if you were to commit that only/separately I'll accept it.

Lseoksee (commented):
@kyrie25 Okay, so I'll leave the second change and the multi-token feature of Popup Lyrics.

kyrie25 (changes requested):
Prettify files.

kyrie25 (approved):

Extensions/popupLyrics.js
```diff
@@ -100,6 +100,8 @@ function PopupLyrics() {
 			const baseURL = `https://apic-desktop.musixmatch.com/ws/1.1/macro.subtitles.get?format=json&namespace=lyrics_synched&subtitle_format=mxm&app_id=web-desktop-app-v1.0&`;
 
 			const durr = info.duration / 1000;
+			const tokens = userConfigs.services.musixmatch.token.split("|");
+			const usertoken = tokens[Math.floor(Math.random() * tokens.length)];
```
rxri, line 104: After discussing with team, this is unnecessary. Lyrics Plus doesn't need two or more tokens to work and doesn't make kinda sense because there is very high change it will pick a token that is "rate-limited" or "doesn't work". I also don't think someone would use it since it would be an undocumented feature. I will implement automatic token generation in a few days instead

rxri (changes requested):

Lseoksee (commented):
@rxri This line has already been a feature in Lyrics Plus. Would you like me to modify the Lyrics Plus as well?

rxri:
please do

Lseoksee:
> please do

I modified it.

rxri (approved):

kyrie25 (approved):

PR #2226, opened 2023-03-29
feat(preprocess): add support for new color vars

ohitstom:
This PR adds support for these color variables:
+ --background-highlight
+ --background-elevated-highlight
+ --background-elevated-base

These are done in form (color.ini):
+ highlight: hex
+ highlight-elevated: hex
+ main-elevated: hex

This is useful as many hover elements currently use Spotifys default hex for these variables, such as history buttons using background-elevated-highlight(#2a2a2a).

And of course if theme developers leave these color.ini variables blank (or not include them at all), they would default back to spotifys hex values, meaning this change would not break anything. 

## What an example color.ini would now look like:
image

## An example of the new highlight/highlight-elevated variables is showcased in these images (using spicetifydefault):
image
image
image
image


ohitstom:
Alot of this is color-theory stuff so this could use a lot of tweaking, but having background-highlight at a minimum available for devs without having to use css would be a huge plus.

kyrie25 (changes requested):
Format code

ohitstom (commented):
done

PR #2227, opened 2023-03-29
feat: add check-update flag

### LINKED ISSUE #2071: Auto update Spicetify

It would be awesome if **spicetify** would update automatically when you start windows / start spotify / there was an update.

### ➕ Additional Information

I've been a few months making use of a script that runs the update commands and have kept that script in the windows startup folder, but it would be great if this was implemented natively in the application itself.

```bash
#!\usr\bin\sh
spicetify upgrade
if [ $? -eq 0 ]; then
    spicetify restore backup apply --yes
    if [ $? -eq 0 ]; then
        echo "Spicetify has been successfully upgraded and restored"
    else
        echo "Spicetify has been successfully upgraded but failed to restore"
        return 1
    fi
else
    echo "Spicetify has failed to upgrade"
    return 1
fi
exit 0
```

rxri:
Although this solution partially resolves #2071, we do not intend to implement a fully-fledged auto-update feature due to the potential risk of Spotify causing disruption to Spicetify's functionality at any point. As a result, our solution involves utilizing the `--check-update` flag, which can be added to any command. This allows users to create a shortcut that can update Spicetify prior to launching Spotify

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -114,10 +114,10 @@ func init() {
 				}
 
 				cmd := exec.Command(ex, cmds...)
-				utils.Scanner(cmd)
+				utils.CmdScanner(cmd)
 
 				cmd = exec.Command(ex, strings.Join(commands[:], " "))
-				utils.Scanner(cmd)
+				utils.CmdScanner(cmd)
 
 				os.Exit(0)
 			}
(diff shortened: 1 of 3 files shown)


kyrie25 (approved):

theRealPadster (approved):

TomSalvador90:
Sorry if it's obvious but that command would be added to spicetify.exe before launch? Would adding that to the config in "spotify_launch_flags" do the job? If not, how would one go about launching Spotify main app through spicetify and making sure that command runs?

PR #2229, opened 2023-03-30
feat: expose `Tippy.js` instance

kyrie25:
Works on Spotify `1.1.90` and later, have not tested below
Expose `Tippy.js` instance used by Spotify, useful for displaying tooltip on vanilla JS elements, namely `Topbar` and `Playbar`, making it more useful to implement `label` and help with accessibility.
Spotify_D4OnggQe0E
Spotify_HzlxnQxg8R

### DIFF
--- a/css-map.json
+++ b/css-map.json
@@ -2165,5 +2165,8 @@
     "eZnAGhYcXE4Bt0a7958z": "main-useDropTarget-playlistV2",
     "ratGUXdpLCkyXZNaJryg": "main-useDropTarget-show",
     "caTDfb6Oj7a5_8jBLUSo": "main-useDropTarget-track",
-    "aRyoyQFJkzhoSOnf2ERM": "main-useDropTarget-local"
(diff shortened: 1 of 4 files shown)


rxri (approved):

Tetrax-10:
dont merge it wait need to change delay variable

Tetrax-10:
Currently in this PR the delay is set default to `500ms` but the default value used by Spotify is `200ms` from `1.1.95.889.gfc9a996d` to present version (`1.2.8.923.g4f94bf0d`). So for being consistant with Spotify, we should use `200ms` as default. Most people are up to date with Spotify version maybe 1 or 2 versions behind.

So existing all extensions will have `500ms` as delay

### 500 ms (not consistent):

https://user-images.githubusercontent.com/75513645/228813783-5b2e3268-de5c-41e8-ba08-70360fcb19ca.mp4

theRealPadster (approved):

PR #2263, opened 2023-04-15
feat: help on no commands

erikhuizinga:
Spicetify v2.17.1 prints the following with no commands:

```
$ spicetify
info Run "spicetify -h" for commands list.
```

It might as well just print the help text that it suggests to the user.

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -134,7 +134,7 @@ func init() {
 	cmd.InitConfig(quiet)
 
 	if len(commands) < 1 {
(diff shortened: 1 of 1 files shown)


Tetrax-10:
Good one 👍. This is the standard for most CLI(s) eg: npm, go, etc...

rxri:
This is not any "standard". We discussed it a little and probably will go with showing available flags and commands without description

kyrie25 (approved):

theRealPadster (approved):

PR #2273, opened 2023-04-18
fix(watch): reload on asset change

### LINKED ISSUE #2271: When updating assets files in assets folder spotify doesnt reload

- [X] I have checked the FAQ

```Markdown
- Spotify version: latest
- Spicetify version:latest
```

### 📝 Description

I have some `CSS` in my assets folder that are appended to the DOM dynamically. when I save them it says `Custom assets are updated` but doesnt say `Spotify reloaded` It would be nice to have auto reload for assets folder.

The files are updated and copied to xpui path but they doesnt trigger reload for some reason.

### 📸 Screenshots

It does says Spotify will auto reload on -s watch for assets

image

_No response_

kyrie25:
Resolves #2271 
Also moved away from deprecated packages

https://user-images.githubusercontent.com/77577746/232745451-d5f7e004-8df5-4e6c-8def-4d785b175884.mp4

### DIFF
--- a/src/utils/watcher.go
+++ b/src/utils/watcher.go
@@ -3,9 +3,10 @@ package utils
 import (
 	"bytes"
 	"encoding/json"
+	"io"
 	"io/fs"
-	"io/ioutil"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -25,7 +26,7 @@ func Watch(fileList []string, callbackEach func(fileName string, err error), cal
 	for {
 		finalCallback := false
 		for _, v := range fileList {
-			curr, err := ioutil.ReadFile(v)
+			curr, err := os.ReadFile(v)
 			if err != nil {
 				callbackEach(v, err)
 				continue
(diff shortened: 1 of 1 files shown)


src/utils/watcher.go
```diff
@@ -3,9 +3,10 @@ package utils
 import (
 	"bytes"
 	"encoding/json"
+	"io"
 	"io/fs"
```
theRealPadster, line 7: Any point importing io and io/fs? Or is it not like JavaScript?

kyrie25: You have to do it this way afaik, `fs` is an extended namespace of `io`

kyrie25: Oh wait nvm, you can get the same type from `os`

kyrie25:
Seems like Go is complaining about the initial `err` parameter not being read before overwriting in the next line.
According to [Golang docs](https://pkg.go.dev/io/fs#WalkDirFunc), the error is non-nil when the directory can't be read, so this should be beneficial.

> WalkDir calls the function with a non-nil err argument in two cases.
>
> First, if the initial fs.Stat on the root directory fails, WalkDir calls the function with path set to root, d set to nil, and err set to the error from fs.Stat.
>
> Second, if a directory's ReadDir method fails, WalkDir calls the function with path set to the directory's path, d set to an fs.DirEntry describing the directory, and err set to the error from ReadDir. In this second case, the function is called twice with the path of the directory: the first call is before the directory read is attempted and has err set to nil, giving the function a chance to return SkipDir or SkipAll and avoid the ReadDir entirely. The second call is after a failed ReadDir and reports the error from ReadDir. (If ReadDir succeeds, there is no second call.)

theRealPadster (approved):

rxri (approved):

PR #2281, opened 2023-04-20
revert(wrapper): notify update in-app

kyrie25:
Turns out this is unnecessary overhead, users will only need to re-apply Spicetify when Spotify itself updates, of which we can enable update checking by default in the CLI. CLI already does a better job at error handling, displaying and updating in the same place also.

Previous commit has also discouraged applying with Spicetify version mismatch so there's really no reason to have 2 update checks in 2 different places.


Tetrax-10:
## Bro really got pissed of LOL

@kyrie25 Dont get emotional Spicetify is not your personal play toy.

image

Using image as an icon which is always white in color is a bad idea. Yeah I did say you `img` sucks its true when using themes. You don't use themes so wont care about other Spicetify users?

kyrie25:
> @kyrie25 Dont get emotional Spicetify is not your personal play toy.

I don't consider it as my personal play toy, and the reason is valid.

Early on it did occur to me that the CLI handles it well, but then it failed to dawn on me that most of the spam issues that were reported on the repository is because people were using outdated Spicetify applied on an incompatible Spotify version.

By already discouraging users from applying on mismatched versions there is no need to run 2 requests to GitHub APIs.

kyrie25:
Also it's my code, I can revoke it however I want.

Tetrax-10:
> Early on it did occur to me that the CLI handles it well, but then it failed to dawn on me that most of the spam issues that were reported on the repository is because people were using outdated Spicetify applied on an incompatible Spotify version.

Spotify Auto updates we can't stop that with `Spicetify` most people just get updated to latest Spotify like recently `1.2.9` where Spicetify broke and the patch was released within 2 days. But still people come to support server and ask for help, they just dont know Spicetify has a update (no user will run `spicetify upgrade in daily basis`), as a Helper all the time I need to say is Update Spicetify with those upgrade commands.

> Requests to GitHub APIs

It doesnt even matter as the APIs are open and there is no API key.

> mismatched versions?

I can fix it just by checking if their current version is listed in the Compatibility section
image

Like If linux users are using `1.1.85` I can stop showing the Spicetify update popup just by checking the condition

Tetrax-10:
> Its my code I can revert it anytime

Lol I just said Spicetify is not your personal play toy.

kyrie25:
> > Its my code I can revert it anytime (deleted message)
> 
> 
> 
> Lol I just said Spicetify is not your personal play toy.

I didn't delete it because it was wrong, but simply not relevant.

Since you're trying to dissuade not only myself but others from closing this unwillingly, I'll have to remind you that it is still within my rights to revoke my work.

Tetrax-10:
> A future version may be supported and not listed in the release note at a later date. This has occurred multiple times for versions that had worked for more than a month or maybe 2.

Bro you do know that we can check if the current spotify is a future version or not right? I said I can fix that If you dont know the logic then just wait.
image
image


> There are rate limits, and even worse for devs the request is fetched every time the client reloads

We dont use Auth Tokens LOL. Also no user is gonna open Spotify 60 times in a hour LOL. You just posting false info so you can revert this commit.
image

> Every time a user tries to re-apply Spotify after a client update the update notification will appear

no user will run `spicetify apply` in daily basis LOL.

# Am not gonna waste time arguing with a Emotionally weak kid who Implemented a feature and now wanna remove that feature because I made a better Icon and said yours `sucks`.

kyrie25:
Rate limits can occur even if you haven't made a request that many times, this has been noted before.

Also lmao.

kyrie25:
That was amusing to watch ngl, but I'm not killing my work that easily.

PR #2284, opened 2023-04-22
fix(lyrics-plus): set icon stroke-width

ohitstom:
Many custom app icons lack their stroke-width being set meaning they all vary in width.
This is a simple PR to fix that.

Example (lyrics-plus):
before:
image

after:
image


ohitstom:
Lmk if it needs formatting and what with, i wasnt sure.

PR #2293, opened 2023-04-25
feat(lyrics-plus): hangul transliteration support

### LINKED ISSUE #2247: [Lyrics Plus] Hangul to Romaja Lyrics Conversion

I hope it can implemented in Lyrics-plus since someone already make the [extension](https://github.com/41pha1/spicetify-extensions/tree/main/romaja-lyrics) for it.
But due the limitation only lyrics-plus can show and integrated at Full Screen Mode, that must be awsome if lyrics-plus also support hangul to romaja conversion like how japanese lyrics do.

### ➕ Additional Information

_No response_

ohitstom:
This PR adds transliteration options for korean synced lyrics (hangul) to roman letters (romaja/latin script), resolves #2247.

image
image


ohitstom:
These are my current possible concerns for code, i believe they should be correct but i dont entirely understand their function and may need correcting/given a proper explanation.

image
image

CustomApps/lyrics-plus/index.js
```diff
@@ -328,6 +335,18 @@ class LyricsContainer extends react.Component {
 				lyricContainerUpdate && lyricContainerUpdate();
 			});
 		});
+
+		[
+			["hangul", "hangul"],
+			["romaja", "romaja"]
+		].map(params => {
```
kyrie25, line 342: Created array has no value and is not needed.
```suggestion
		].forEach(params => {
```

CustomApps/lyrics-plus/Translator.js
```diff
@@ -78,6 +88,19 @@ class Translator {
 		});
 	}
 
+	async convertToRomaja(text, target) {
+		if (!this.finished) {
+			setTimeout(this.convertToRomaja.bind(this), 100, text, mode);
+			return;
+		}
+
+		if (target == "hangul") {
```
kyrie25, line 97:
```suggestion
		if (target === "hangul") {
```

kyrie25 (changes requested):

kyrie25 (changes requested):

ohitstom:
closed and reopened cause flow wouldnt initialize

CustomApps/lyrics-plus/Utils.js
```diff
@@ -77,6 +77,9 @@ class Utils {
 			if (character.match(kanaRegex)) {
 				kanaCount++;
 			}
+			if (character.match(hangulRegex)) {
+				return "ko";
+			}
```
kyrie25, line 82: Place these above the other conditions so it doesn't execute unnecessary code.

kyrie25 (changes requested):

CustomApps/lyrics-plus/Translator.js
```diff
@@ -78,6 +88,19 @@ class Translator {
...
+		if (!this.finished) {
+			setTimeout(this.convertToRomaja.bind(this), 100, text, target);
+			return;
+		}
+
+		if (target === "hangul") {
+			return text;
+		} else {
+			return Aromanize.hangulToLatin(text, "rr-translit");
+		}
```
kyrie25, line 101:
```suggestion
		if (target === "hangul") return text;
		return Aromanize.hangulToLatin(text, "rr-translit");
```

kyrie25 (changes requested):
Last one.

kyrie25 (approved):

rxri (changes requested):
I don't think it's necessary to load every library when it's not needed, shouldn't we load them accordingly to what user enables?

Tetrax-10:
> I don't think it's necessary to load every library when it's not needed, shouldn't we load them accordingly to what user enables?

#2122 in this PR the translation is only enabled only when the user clicks a translation option. The same you can just check what translation option is selected and then load that minified Library.

rxri (approved):
after some discussion I came to conclusion that it can be done later on

PR #2295, opened 2023-04-26
fix(exp-features): save config after reset & add fallback method

kyrie25:
Save configuration properly instead of temporary changing value in features map. This resolves the issue of configuration resolving to local values instead of remote after a reset and a reload.

Also refactor some parts.


jsHelper/expFeatures.js
```diff
@@ -203,61 +203,54 @@
...
+		search.oninput = () => {
+			const query = search.value.toLowerCase();
+			const rows = content.querySelectorAll(".setting-row");
+			rows.forEach(row => {
+				if (row.id === "search" || row.id === "reset") return;
+				if (row.textContent.trim().toLowerCase().includes(query) || row.id.toLowerCase().includes(query)) {
+					row.style.display = "flex";
+				} else {
+					row.style.display = "none";
+				}
```
rxri, line 230:
```suggestion
				row.style.display = row.textContent.trim().toLowerCase().includes(query) || row.id.toLowerCase().includes(query) ? "flex" : "none"
```

rxri (changes requested):

Tetrax-10:
Yup this one works fine. The problem occurs when resetting the exp features.

Like for some people Library x was enabled remotely. But when resetting the exp features the lib x was applied but then after a refresh the UI reverts to old UI. After this the resting works fine.

rxri (approved):

kyrie25:
For some reason a small amount of people are getting old versions of `xpui` so added a fallback method just in case it is an old version or if the hook breaks in the future.
IZP3hrsdmv

theRealPadster (approved):

PR #2304, opened 2023-04-29
fix(lyrics-plus): handle Musixmatch ratelimit

kyrie25:
Happens after a few reloads, would throw lyrics fetching for the whole session

### DIFF
--- a/CustomApps/lyrics-plus/ProviderMusixmatch.js
+++ b/CustomApps/lyrics-plus/ProviderMusixmatch.js
@@ -116,7 +116,7 @@ const ProviderMusixmatch = (function () {
 		if (isInstrumental) {
 			return [{ text: "♪ Instrumental ♪", startTime: "0000" }];
 		} else if (hasSynced) {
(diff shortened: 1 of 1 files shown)


CustomApps/lyrics-plus/index.js
```diff
@@ -224,25 +224,53 @@ class LyricsContainer extends react.Component {
...
+			}
+
+			if (!data[currentMode]) {
+				for (const key in data) {
+					if (data[key] && !finalData[key]) {
+						finalData[key] = data[key];
+					}
+				}
+				continue;
+			}
```
rxri, line 256:
```suggestion
			if (data[currentMode]) continue;
			for (const key in data) {
				if (data[key] && !finalData[key]) {
					finalData[key] = data[key];
				}
			}
```

rxri (changes requested):

rxri (approved):
don't mind my previous review

kyrie25:
Will attempt to use GraphQL extracted colors (the same one used for Now Playing View and gradients) for Lyrics Plus, while keeping old method as fallback.
image

theRealPadster (approved):

PR #2305, opened 2023-04-30
feat: hook GraphQL wrapper

kyrie25:
Hooks onto Spotify's GraphQL handler and definitions. This allows developers to make queries to Spotify GraphQL endpoints easily without having to use obscure hashes for persisted query validation.
Works on `1.1.90` and above, have not tested below.
image
image

This also includes all possible definitions for accepted queries
image
image
These definitions can be passed to the `Request` method as the first parameter for each query you want to fetch.
Additionally, since these are GraphQL definitions, they also include parameter type definitions so developers can easily catch on and utilize it.
image

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -1183,4 +1183,38 @@ declare namespace Spicetify {
          */
         function sub(callback: (title: string) => void): { clear: () => void };
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


rxri (approved):

theRealPadster (approved):

rxri (approved):

rxri:
double review 🔥

PR #2310, opened 2023-05-01
feat(lyrics-plus): improve Netease lyrics algorithm

Lseoksee:
1. The existing Netease lyrics were too inaccurate. In addition to comparing album names, this commit adds a method of comparing the length of the song in milliseconds (milliseconds are specified as ±500 milliseconds because there is a slight error), if either of these conditions is not met, it returns no lyrics (because otherwise useless lyrics will be returned)
2. The return value "纯音乐, 请欣赏" for lyrics Netease doesn't seem to be the track Instrumental. Sometimes when listening to local music or songs by indie artists, Instrumental is displayed. Therefore, it was modified to be displayed as no lyrics.
3. Fixed an issue where, when viewing the lyrics of Netease" in popupLyrics, the song would be displayed as prelude if there was a blank during the lyrics. the problem did not exist in lyrics-plus, so please leave an answer if this feature is intended.

### DIFF
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -121,7 +121,7 @@ const ProviderNetease = (function () {
 		const lyrics = lines
 			.map(line => {
 				const { time, text } = parseTimestamp(line);
-				if (text === "纯音乐, 请欣赏") nolyr = true;
+				if (text === "纯音乐, 请欣赏") noLyrics = true;
 				if (!time || !text) return null;
 
 				const [key, value] = time.split(":") || [];
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -111,7 +111,7 @@ const ProviderNetease = (function () {
 
 	function getSynced(list) {
 		const lyricStr = list?.lrc?.lyric;
-		let nolyr = false;
+		let noLyrics = false;
 
 		if (!lyricStr) {
 			return null;
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -175,7 +175,7 @@ const ProviderNetease = (function () {
 
 	function getUnsynced(list) {
 		const lyricStr = list?.lrc?.lyric;
-		let nolyr = false;
+		let noLyrics = false;
 
 		if (!lyricStr) {
 			return null;
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -17,8 +17,8 @@ const ProviderNetease = (function () {
 		}
 
 		const album = Utils.capitalize(info.album);
-		let itemId = items.findIndex(val => Utils.capitalize(val.album.name) === album);
-		if (itemId === -1) itemId = 0;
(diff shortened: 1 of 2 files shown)


kyrie25:
> 1. The existing Netease lyrics were too inaccurate. In addition to comparing album names, this commit adds a method of comparing the length of the song in milliseconds (milliseconds are specified as ±500 milliseconds because there is a slight error)

I think the error margin can be increased, as there can be instances where you'd need to increase up to 2000-3000ms at most for offset. I'm assuming that should be safe, though can you give an opinion on that suggestion or test it out?

Lseoksee:
> I think the error margin can be increased, as there can be instances where you'd need to change up to 2000-3000ms at most for offset. I'm assuming that should be safe, though can you give an opinion on that suggestion or test it out?

Tests have shown that when offset is over 1000, a small number of songs bring lyrics incorrectly. So I think 1000 is the maximum limit also, when I tested offset at 500, I rarely went out of the offset range, so I think it would be good to set it to 1000.

kyrie25:
> Tests have shown that when offset is over 1000, a small number of songs bring lyrics incorrectly. So I think 1000 is the maximum limit also, when I tested offset at 500, I rarely went out of the offset range, so I think it would be good to set it to 1000.

Sounds good to me.

rxri (changes requested):
Please also run prettier

Lseoksee:
> Please also run prettier

Why does the "prettier" error occur?

Lseoksee:
@kyrie25  @rxri   `Fatal: could not read Username for 'https://github.com ': No Such device or address` error occurs during prettier operation, how do I resolve it?

rxri:
Run prettier in your vscode with `.prettierrc` found in our repo to prettify commands. Don't mind Action throwing this error because it's trying to push commit that has it prettified but it can't

Lseoksee:
I solved it!

rxri (approved):

CustomApps/lyrics-plus/ProviderNetease.js
```diff
@@ -17,8 +17,10 @@ const ProviderNetease = (function () {
 		}
 
 		const album = Utils.capitalize(info.album);
-		let itemId = items.findIndex(val => Utils.capitalize(val.album.name) === album);
-		if (itemId === -1) itemId = 0;
+		let itemId = items.findIndex(
+			val => Utils.capitalize(val.album.name) === album || (info.duration + 1000 > val.duration && info.duration - 1000 < val.duration)
```
kyrie25, line 21: This should be valid.
```suggestion
			val => Utils.capitalize(val.album.name) === album || Math.abs(info.duration - val.duration) < 1000
```

Extensions/popupLyrics.js
```diff
@@ -175,8 +175,10 @@ function PopupLyrics() {
 			}
 
 			const album = LyricUtils.capitalize(info.album);
-			let itemId = items.findIndex(val => LyricUtils.capitalize(val.album.name) === album);
-			if (itemId === -1) itemId = 0;
+			let itemId = items.findIndex(
+				val => LyricUtils.capitalize(val.album.name) === album || (info.duration + 1000 > val.duration && info.duration - 1000 < val.duration)
```
kyrie25, line 179:
```suggestion
				val => LyricUtils.capitalize(val.album.name) === album || Math.abs(info.duration - val.duration) < 1000
```

kyrie25 (changes requested):

kyrie25 (approved):

PR #2319, opened 2023-05-03
feat: modify lyrics cleantitle regex & netease lyrics update

Lseoksee:
In addition to `feat` and `with`, we modified the regular expression to detect `prod`, which means the producer of music.
Additionally, when empty lyrics are delivered in popupLyrics, there is an empty array, and the empty lyrics are printed out, so we modified it to print out that there are no lyrics.


Lseoksee:
Additionally, the regular expression has been updated to prevent the mix metadata from being loaded when loading the Netease lyrics. In addition, we modified the regular expression of lirics-plus and the regular expression of popupLyrics equally.

Lseoksee:
I changed the notation of Interlude in the Musixmatch lyrics of popupLyrics from `...` to `♪`

kyrie25 (approved):
LGTM

rxri (approved):

PR #2328, opened 2023-05-06
feat(new-releases): dismissible cards

### LINKED ISSUE #2317: Hide releases from "New Releases" tab

I really like the "new releases" tab, it has made following artists very convenient! It would be useful to have e.g. a cross icon/hide button for each release to remove it from the list, for releases I've already listened to/bought elsewhere/am not interested in, to keep the calendar clean and easy to use (i have 2k followed artists, so it gets busy!). I made a (very ugly) mockup with two versions of what I'm imagining the feature to look like.

Thanks for your consideration :)

### ➕ Additional Information

image

ohitstom:
Adds a close button to the top right of cards, once clicked a function is ran that adds the releases uri to an array which is then used by refresh() to refresh the grid whilst skipping those specific uris, aims to resolve #2317.

image

example video:
https://cdn.discordapp.com/attachments/842219447716151310/1104487178702180423/screen-capture.webm

### DIFF
--- a/CustomApps/new-releases/Card.js
+++ b/CustomApps/new-releases/Card.js
@@ -21,6 +21,7 @@ class Card extends react.Component {
 
 	closeButtonClicked(event) {
 		removeCards(this.props.uri);
+		Spicetify.showNotification(`Dismissed <b>${this.title}</b> from <b>${this.artist.name}</b>`);
 		event.stopPropagation();
 	}
 
@@ -112,7 +113,7 @@ class Card extends react.Component {
 						),
 						react.createElement(
 							Spicetify.ReactComponent.TooltipWrapper,
-							{ label: "Close" },
+							{ label: "Dismiss" },
 							react.createElement(
 								"button",
 								{
--- a/CustomApps/new-releases/Card.js
+++ b/CustomApps/new-releases/Card.js
@@ -19,6 +19,11 @@ class Card extends react.Component {
 		event.stopPropagation();
 	}
 
+	closeButtonClicked(event) {
+		removeCards(this.props.uri);
+		event.stopPropagation();
+	}
+
 	render() {
 		let detail = [];
 		this.visual.type && detail.push(this.type);
@@ -44,7 +49,7 @@ class Card extends react.Component {
 					"div",
 					{
 						className: "main-card-draggable",
-						draggable: "true"
+						draggable: "false"
 					},
 					react.createElement(
 						"div",
@@ -103,6 +108,34 @@ class Card extends react.Component {
 										)
 									)
 								)
+							)
+						),
+						react.createElement(
+							"button",
+							{
+								className: "main-card-closeButton",
+								onClick: this.closeButtonClicked.bind(this),
+								style: {
+									position: "absolute",
+									top: "10px",
+									right: "10px"
+								}
+							},
+							react.createElement(
+								"svg",
+								{
+									width: "18",
+									height: "18",
+									viewBox: "0 0 32 32",
+									xmlns: "http://www.w3.org/2000/svg",
+									className: "main-card-closeButton-svg"
+								},
+								react.createElement("title", null, "Close"),
+								react.createElement("path", {
+									d: "M31.098 29.794L16.955 15.65 31.097 1.51 29.683.093 15.54 14.237 1.4.094-.016 1.508 14.126 15.65-.016 29.795l1.414 1.414L15.54 17.065l14.144 14.143",
+									fill: "var(--spice-text)",
+									fillRule: "evenodd"
+								})
 							)
 						)
 					),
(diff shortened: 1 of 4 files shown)


ohitstom:
@kyrie25 would love input on this please.

CustomApps/new-releases/index.js
```diff
@@ -40,10 +40,12 @@ const CONFIG = {
 	relative: getConfig("new-releases:relative", false)
 };
 
+let dismissed = [JSON.parse(Spicetify.LocalStorage.get("new-releases:dismissed")) || null].flat();
```
kyrie25, line 43: Catch `JSON.parse` in case it throws an error, like in any other part of our codebase. Also since if it is already an array, you won't need to construct a new array out of it.
```suggestion
let dismissed;
try {
	dismissed = JSON.parse(Spicetify.LocalStorage.get("new-releases:dismissed");
	if (!Array.isArray(dismissed)) throw "";
} catch {
	dismissed = [];
}
```

CustomApps/new-releases/index.js
```diff
@@ -100,7 +102,13 @@ class Grid extends react.Component {
 		this.setState({ cards: [...gridList] });
 	}
 
-	async reload() {
+	removeCards(id) {
+		dismissed = dismissed[0] === null ? [id] : dismissed.concat(id);
```
kyrie25, line 106: Array variables are mutable.
```suggestion
		dismissed.push(id);
```

CustomApps/new-releases/index.js
```diff
@@ -130,6 +138,10 @@ class Grid extends react.Component {
 		}
 
 		for (const track of items) {
+			if (dismissed.includes(track.uri)) {
+				continue;
+			}
```
kyrie25, line 143:
```suggestion
			if (dismissed.includes(track.uri)) continue;
```

CustomApps/new-releases/index.js
```diff
@@ -300,9 +313,11 @@ var count = (function () {
 	};
 })();
 
-async function fetchTracks() {
+async function fetchTracks(silent) {
 	let artistList = await getArtistList();
-	Spicetify.showNotification(`Fetching releases from ${artistList.length} artists`);
+	if (!silent) {
+		Spicetify.showNotification(`Fetching releases from ${artistList.length} artists`);
+	}
```
kyrie25, line 320:
```suggestion
	if (!silent) Spicetify.showNotification(`Fetching releases from ${artistList.length} artists`);
```

kyrie25 (changes requested):

CustomApps/new-releases/Card.js
```diff
@@ -44,7 +49,7 @@ class Card extends react.Component {
 					"div",
 					{
 						className: "main-card-draggable",
-						draggable: "true"
+						draggable: "false"
```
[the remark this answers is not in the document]
ohitstom: The cards are draggable in a way they shouldnt be, i believe this was an oversite or a mistype since all the over card divs have draggable false.

CustomApps/new-releases/Card.js
```diff
@@ -104,6 +109,34 @@ class Card extends react.Component {
...
+						react.createElement(
+							"button",
+							{
+								className: "main-card-closeButton",
+								onClick: this.closeButtonClicked.bind(this),
+								style: {
+									position: "absolute",
+									top: "10px",
+									right: "10px"
+								}
```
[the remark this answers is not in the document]
ohitstom: will do 👍🏻

CustomApps/new-releases/index.js
ohitstom: this had a reason since it would append after a null value if the list was empty, could be an oversight in how i was fetching localstorage however and your suggestions above could fix it.

CustomApps/new-releases/Card.js
```diff
@@ -19,6 +19,11 @@ class Card extends react.Component {
 		event.stopPropagation();
 	}
 
+	closeButtonClicked(event) {
+		removeCards(this.props.uri);
+		event.stopPropagation();
```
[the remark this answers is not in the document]
ohitstom: why wouldnt it be? i need to pass the uri to the index and stop the event afterwards, this also means if it errors it just opens the cards page instead of crashing spotify.

kyrie25: > this also means if it errors it just opens the cards page instead of crashing spotify.

Execution halts when an exception is thrown, so that doesn't help with anything.

kyrie25: We already established that this was meant to not open the page after clicking the dismiss button.

kyrie25 (approved):

kyrie25 (changes requested):

kyrie25 (approved):

rxri (approved):

PR #2330, opened 2023-05-07
fix(lyrics-plus): seperate translated lyrics providers from mode selection

Deskehs:
This was supposed to allow Netease translated lyrics covert to Traditional Chinese.

However, in the future if Musixmatch translated lyrics are supported, they can also (for example in Japanese) convert to furigana.

This PR also updates the language detection algorithm a bit, allowing detection override and some fixes and one-liner changes.

NOT COMPLETE: just need to figure how to run the translate function and dropdown update after change


CustomApps/lyrics-plus/Utils.js
```diff
@@ -67,33 +69,23 @@ class Utils {
...
-				tradCount++;
-			}
-		}
+		const kanaCount = cjkMatch.filter(glyph => kanaRegex.test(glyph)).length;
+		const simpCount = cjkMatch.filter(glyph => simpRegex.test(glyph)).length;
+		const tradCount = cjkMatch.filter(glyph => tradRegex.test(glyph)).length;
+		const hangulCount = cjkMatch.filter(glyph => hangulRegex.test(glyph)).length;
 
 		const kanaPercentage = kanaCount / cjkMatch.length;
+		const hangulPercentage = hangulCount / cjkMatch.length;
```
ohitstom, line 81: You dont need to do hangul percentage since its regex will either be true or false.

Deskehs: I purposefully want to count the percentage of the Hangul included of the lyrics because of a possibility that a Chinese/Japanese song which includes some Korean references in which the lyrics show it in Korean. I do not want to make the whole song mistakenly detected as Korean while it is Chinese/Japanese. Sure, it's a hypothetical situation but don't disregard it either.

ohitstom: there are no Korean characters in either the Japanese or Chinese alphabet so this is simply unneeded, kyrie had already told me not to do this in one of my previous PRs.

Deskehs: Oh. In the case because of Kyrie said so I guess I have to retract this.

Deskehs:
I kind of need some help here. As I don't have much React experience I do not know how would I switch state between the translated lyrics from the provider and the page's original lyrics without overriding them.

CustomApps/lyrics-plus/Utils.js
Deskehs: I will revert this tomorrow.

PR #2344, opened 2023-05-11
feat(fullAppDisplay): add option to display release date

Ender-Wang:
- Previously, releaseDate is appended to the album name by `albumText += " • " + (await this.getAlbumDate(albumURI));`. Sometimes if the album name is a little bit longer, some part of the releaseDate string will be pushed to the next line, mostly it's the year, but sometimes even the month or the whole releaseDate string), which does not look good. 
- Now, everything matches the rest, just added an option to toggle releaseDate on and off🥳🥳🥳!

### DIFF
--- a/Extensions/fullAppDisplay.js
+++ b/Extensions/fullAppDisplay.js
@@ -132,7 +132,7 @@ body.video-full-screen.video-full-screen--hide-ui {
     color: currentColor;
     padding: 0 5px;
 }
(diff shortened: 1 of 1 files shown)


Ender-Wang:
Changeset preview:
image

kyrie25 (approved):

rxri (approved):

PR #2348, opened 2023-05-14
style(new-releases): match spotify card styling

ohitstom:
Comparison of the new border radius of new-releases card (blue), and default spotify card (red).
Untitled-1

Since we are unable to receive any of the crucial variables that the radius relies on, i have taken advantage of the fact clamp will default to 0px when --column width isnt present, and since this variable isnt relied on anywhere else there is no need for it and will default to 0px in the clamp as spotify intends.
image

This is a pretty janky solution to come to, but it was either that or modify the --card-container-border-radius mentioned above for a multitude of classes.

this leaves us looking like this:
image

### DIFF
--- a/CustomApps/new-releases/Card.js
+++ b/CustomApps/new-releases/Card.js
@@ -123,14 +123,14 @@ class Card extends react.Component {
 								react.createElement(
 									"svg",
 									{
-										width: "18",
-										height: "18",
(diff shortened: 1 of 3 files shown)


rxri (approved):

kyrie25 (approved):
LGTM

PR #2363, opened 2023-05-20
chore(types): update URI class definition

Aimarekin:
URIs can have a track property for local tracks. For example,
```js
Spicetify.URI.from("spotify:local:Andrew+Hale+%26+Simon+Hale:L.A.+Noire:Global+Chase+Full:129")
```

returns 
```js
{
    "type": "local",
    "artist": "Andrew Hale & Simon Hale",
    "album": "L.A. Noire",
    "track": "Global Chase Full",
    "duration": 129
}
```

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -474,6 +474,7 @@ declare namespace Spicetify {
         public args?: any;
         public category?: string;
         public username?: string;
+        public track?: string;
         public artist?: string;
         public album?: string;
         public query?: string;


kyrie25 (commented):
Can you add `duration` to the list of public properties while you're at it?

Aimarekin:
> Can you add `duration` to the list of public properties while you're at it?

Right, didn't notice - I'll make a new PR

PR #2367, opened 2023-05-21
fix(trashbin): disable widget on song change

kyrie25:
Fix bug where widget appears after song change on disabled


CharlieS1103 (approved):
LGTM

PR #2390, opened 2023-06-02
feat(trashbin): handle current track upon enable.

Qwerty-133:
By invoking watchChange when the event listeners are activated, we can handle tracks that are already playing upon the initial loading of the extension, as well as tracks that are playing when the extension has just been enabled through the settings page.


kyrie25 (changes requested):
Merge this with your previous PR and change the scope

kyrie25:
Superseded by #2389

Issue #2394, opened 2023-06-03
Adding assets/resources to custom apps


kuba2k2:
### 📝 Provide a description of the new feature

I'm writing a custom app that needs to access its own resources, stored in files (like JSONs or images). I can see that there are options to include `subfiles` and `subfiles_extension` in `manifest.json`, however these are only limited to JS files. I haven't found a way to include other resources. Of course, one could bundle them inside .js files (with something like webpack) but that complicates retrieving them and will *greatly* increase the JS size (over 30 MiB in my case).

### ➕ Additional Information

[Here's the pushApps function](https://github.com/spicetify/spicetify-cli/blob/master/src/cmd/apply.go#L331), which could (similarly to subfiles) copy assets, specified in the manifest, to the xpui directory (most likely, a subdirectory like `assets/<my-app-name>`).

kuba2k2:
Added in a PR #2396.

PR #2396, opened 2023-06-04
feat(custom-apps): allow adding assets/resources

kuba2k2:
#2394 

Example usage (manifest.json):
```json
{
  "name": "...",
  "subfiles": ["..."],
  "assets": [
    "manifest.json",
    "beatsaber.*.js",
    "css/*.css",
    "levelpacks",
    "levels/Ost*.png"
  ]
}
```

Glob expressions are supported as well (`css/*.css`). Directories can also be specified by just their name/path (`levelpacks`).

File/dir structure is resembled in the output directory.

I can add this to docs with a separate PR if requested.

### DIFF
--- a/src/cmd/apply.go
+++ b/src/cmd/apply.go
@@ -344,6 +344,36 @@ func pushApps(list ...string) {
 				}
 				pushExtensions(app, subfilePath)
 			}
+			for _, assetFile := range manifestJson.Assets {
+				assetsList, err := filepath.Glob(filepath.Join(customAppPath, assetFile))
+				if err != nil {
+					utils.PrintError(err.Error())
+					continue
+				}
+				for _, assetPath := range assetsList {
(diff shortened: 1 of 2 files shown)


src/cmd/apply.go
```diff
@@ -344,6 +344,36 @@ func pushApps(list ...string) {
 				}
 				pushExtensions(app, subfilePath)
 			}
+			for _, assetFile := range manifestJson.Assets {
+				assetsList, err := filepath.Glob(filepath.Join(customAppPath, assetFile))
```
kyrie25, line 348: We should add another case here where `assetList` returns an empty array, so we can rule out false positives when reporting issues.

Something like printing out a warning when length is 0.

kyrie25 (changes requested):

kyrie25 (approved):
LGTM

PR #2403, opened 2023-06-11
feat(watch): enable devtools

machinemessiah:
For example `spicetify watch -[insert optional flag(s) here]` will watch for changes and restart as expected but with the devtools enabled so a separate command is not needed to enable devtools. This is useful for debugging and development.

Works just fine, though as far as I can tell this is certainly not the most efficient way to do this, so someone with a better understanding of the codebase should probably take a look at this.

### DIFF
--- a/src/cmd/watch.go
+++ b/src/cmd/watch.go
@@ -221,6 +221,7 @@ func startDebugger() {
 	if len(utils.GetDebuggerPath()) == 0 {
 		EvalSpotifyRestart(true, "--remote-debugging-port=9222", "--remote-allow-origins=*")
 		utils.PrintInfo("Spotify is restarted with debugger on. Waiting...")
+		SetDevTools();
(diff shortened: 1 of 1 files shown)


machinemessiah:
Apologies for all the edits, I haven't made a pull request in a long time and it took me a minute to work out why the Lint check was failing (hopefully my PR title is appropriate now).

I honestly don't know what the `Build` check requirements are, so I don't know why it's failing; I'm not very familiar with Go as a language, so all I know is that I can build and run the project just fine locally with the small change I made (via `go build` and then running `spicetify-cli.exe watch -l`). Could someone explain what I'm missing here? (different version of Go? I'm running 1.20.5 for reference)

kyrie25 (approved):

theRealPadster (approved):

rxri (approved):

PR #2409, opened 2023-06-14
feat(panel): pass panel ID as props

kyrie25:
Pass panel ID as props since the Panel header requires the panel ID to be passed (doesn't seem to be optional internally, but still works fine without?), plus for more usage inside the panel component as the ID is not accessible until after the panel is registered.
Works fine with both types of children.
image
image

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -1713,7 +1713,7 @@ Spicetify.Playbar = (function() {
                     )
                 )
 
-            contentMap.set(id, Spicetify.React.createElement(ErrorBoundary, { id }, Content));
+            contentMap.set(id, Spicetify.React.createElement(ErrorBoundary, { id }, content));
 
             let isActive = Spicetify.Panel.currentPanel === id;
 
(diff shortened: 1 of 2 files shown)


theRealPadster (approved):

rxri (approved):

jsHelper/spicetifyWrapper.js
```diff
@@ -1681,22 +1682,22 @@ Spicetify.Playbar = (function() {
         subPanelState: (callback) => Spicetify.Platform.PanelAPI.subscribeToPanelState(callback),
         registerPanel: ({ label, children, isCustom = false, style, wrapperClassname, headerClassname, headerVariant, headerSemanticColor, headerLink, headerActions, headerOnClose, headerPreventDefaultClose, headerOnBack }) => {
             const id = [...contentMap.keys()].sort((a, b) => a - b).pop() + 1;
-            const content = isCustom
+            const Content = isCustom
```
rxri, line 1685:
```suggestion
            const content = isCustom
```

jsHelper/spicetifyWrapper.js
```diff
@@ -1708,11 +1709,11 @@ Spicetify.Playbar = (function() {
                             titleVariant: headerVariant,
                             titleSemanticColor: headerSemanticColor,
                         }),
-                        children
+                        Spicetify.React.cloneElement(children, { panel: id })
                     )
                 )
 
-            contentMap.set(id, Spicetify.React.createElement(ErrorBoundary, { id }, content));
+            contentMap.set(id, Spicetify.React.createElement(ErrorBoundary, { id }, Content));
```
rxri, line 1716:
```suggestion
            contentMap.set(id, Spicetify.React.createElement(ErrorBoundary, { id }, content));
```

PR #2426, opened 2023-06-21
feat(wrapper): expose `ReactQuery` & `usePanelState`

kyrie25:
Not much to say except moving to Webpack hooking for `DragHandler` instead of maintaining 2 dreadful hooks
Cool stuff also
image

### DIFF
--- a/css-map.json
+++ b/css-map.json
@@ -2498,5 +2498,6 @@
     "YUyFyiI58gL8VuLdbOD6": "main-devicePicker-troubleshooting",
     "AN2rvWrkrs7UsnY12hL8": "main-devicePicker-troubleshootingList",
     "QClIatTm05fvjVzODr8X": "main-devicePicker-troubleshootingItemIcon",
-    "OTuSSciCS3NJqFG8OKX2": "main-devicePicker-troubleshootingItemSubtitle"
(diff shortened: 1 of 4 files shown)


rxri (approved):

css-map.json
```diff
@@ -2498,5 +2498,6 @@
     "YUyFyiI58gL8VuLdbOD6": "main-devicePicker-troubleshooting",
     "AN2rvWrkrs7UsnY12hL8": "main-devicePicker-troubleshootingList",
     "QClIatTm05fvjVzODr8X": "main-devicePicker-troubleshootingItemIcon",
-    "OTuSSciCS3NJqFG8OKX2": "main-devicePicker-troubleshootingItemSubtitle"
+    "OTuSSciCS3NJqFG8OKX2": "main-devicePicker-troubleshootingItemSubtitle",
+    "SFgYidQmrqrFEVh65Zrg": "main-userWidget-boxCondensed"
```
afonsojramos, line 2502: Can we put this next to the other `userWidget`s? Or does it not make sense? If we are assuming that the CSS map is serving more than one version, then we should try to group them at least, no?

afonsojramos (approved):

css-map.json
kyrie25: I don’t think it matters either way but if you think that’d look nicer.

PR #2455, opened 2023-07-02
feat(install): prevent running as admin or root

rxri:


install.ps1
```diff
@@ -86,6 +86,11 @@ function Move-ConfigFolder {
 
 #region Main
 if ($PSVersionTable.PSVersion.Major -ge $PSMinVersion) {
+  if ([bool]([Security.Principal.WindowsIdentity]::GetCurrent().Groups -contains 'S-1-5-32-544')) {
+    Write-Log -ActionText "The script was ran as Administrator. Exiting..." | Out-Null
```
kyrie25, line 90: We should probably add another line telling users NOT to do that instead of leaving them confused imo.

kyrie25 (approved):

afonsojramos (approved):

PR #2470, opened 2023-07-13
chore(config.go): remove `PrintSuccess` on config generate

SunsetTechuila:
so that automated tools (such as scripts) can rely on the command output


kyrie25:
it is an essential part of user experience, and also of the debugging and development process. Under no circumstance should changes made to the filesystem be left unnotified.

PR #2475, opened 2023-07-16
feat(wrapper): expose `imageColorExtractor` and `Color`

Aimarekin:
Exposes `Spicetify.imageColorExtractor`, which uses Spotify's algorithms to extract the prominent colors from an image URL. Internally, it first attempts to fetch these colors through GraphQL. If it fails, it loads the image and processes it locally. This makes it work for any image, including local tracks:
image
image
or furry art:
image
but not normal item URIs (they are not images):
image
Spotify returns rgb(83,83,83) as fallback.

The colors are returned as Spotify's own internal `Color` class, which is also exposed through this PR.

`colorExtractor` should remain and keep it's original functionality (as it and `imageColorExtractor` expect different inputs, and return different outputs).

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -363,6 +363,20 @@ declare namespace Spicetify {
      */
     function getAudioData(uri?: string): Promise<any>;
     /**
+     * Fetch extracted colors from an image. Can be a URL, or a URI string to an image. Works for both local and remote images.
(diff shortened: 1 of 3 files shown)


Aimarekin:
Compatibility issues with spotify 1.2.13

PR #2501, opened 2023-08-04
fix(css-map): remove outdated buddy feed header class

SunsetTechuila:
This would fix misaligned buddy feed header

Before version 1.2.0 buddy feed header has class `PJt6DU4NhGjz7fVCZjS7` and its styles were stored in `xpui-routes-buddy-feed.css`, but starting with version 1.2.0 the class name has changed and its styles moved to `xpui.css`, so the styles in `xpui-routes-buddy-feed.css` became obsolete and weren't apply anymore. Spicetify changes both the actual class name in `xpui.css` and the obsolete ones in `xpui-routes-buddy-feed.css` to `main-buddyFeed-header`, resulting in incorrect styling and misalignment

stock:
image

with spicetify applied:

image

since versions below 1.2.0 aren't supported anymore and in 1.2.0 it is different class name, I think this class can be removed from the map

image


kyrie25 (approved):
LGTM

rxri (approved):

PR #2506, opened 2023-08-08
refactor(keyboardShortcut): cleanup & patch keybinds

ohitstom:
This PR removes redundant keybinds that Spotify have implemented into their own shortcut menu (ctrl + ?), whilst also fixing pre-existing ones (namely rotate navbar item which broke from libx).

The registering of these binds has also been moved to a json/object structure making it easier to expand upon both on the dev side and in a future settings menu if we were to add one.

Also moved from `Spicetify.Keyboard` to `Spicetify.Mousetrap`

(docs will need rewriting)

### DIFF
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -12,7 +12,6 @@
 
 	// Default keybinds
 	const vim = new VimBind();
-	vim.setCancelKey("ESCAPE");
 	const SCROLL_STEP = 25;
 	const binds = {
 		// Ctrl + Tab and Ctrl + Shift + Tab to switch sidebar items
@@ -77,6 +76,7 @@
 		f: {
 			callback: event => {
 				vim.activate(event);
+				vim.setCancelKey("ESCAPE");
 			}
 		}
 	};
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -49,7 +49,6 @@
 		if (typeof staticCondition === "undefined" || staticCondition) {
 			Spicetify.Mousetrap.bind(key, event => {
 				if (!vim.isActive) {
-					console.log(callback)
 					callback(event);
 				}
 			});
--- a/Extensions/keyboardShortcut.js
+++ b/Extensions/keyboardShortcut.js
@@ -14,65 +14,14 @@
 	const vim = new VimBind();
 	const SCROLL_STEP = 25;
 	const binds = {
-		// Ctrl + Tab and Ctrl + Shift + Tab to switch sidebar items
-		"ctrl+tab": {
-			callback: () => {
-				rotateSidebar(1);
-			}
-		},
-		"ctrl+shift+tab": {
-			callback: () => {
-				rotateSidebar(-1);
-			}
-		},
-		// PageUp, PageDown to focus on iframe app before scrolling
-		"shift+pageup": {
-			callback: focusOnApp
-		},
-		"shift+pagedown": {
-			callback: focusOnApp
-		},
-		// J and K to vertically scroll app
-		j: {
-			callback: () => {
-				const app = focusOnApp();
-				if (app) {
-					const scrollInterval = setInterval(() => {
-						app.scrollTop += SCROLL_STEP;
-					}, 10);
-					document.addEventListener("keyup", () => {
-						clearInterval(scrollInterval);
-					});
-				}
-			}
-		},
-		k: {
-			callback: () => {
-				const app = focusOnApp();
-				if (app) {
-					const scrollInterval = setInterval(() => {
-						app.scrollTop -= SCROLL_STEP;
-					}, 10);
-					document.addEventListener("keyup", () => {
-						clearInterval(scrollInterval);
-					});
-				}
-			}
-		},
-		// G and Shift + G to scroll to top and to bottom
-		g: {
-			callback: () => {
-				const app = focusOnApp();
-				app.scroll(0, 0);
-			}
-		},
-		"shift+g": {
-			callback: () => {
-				const app = focusOnApp();
-				app.scroll(0, app.scrollHeight);
-			}
-		},
-		// F to activate Link Follow function
+		"ctrl+tab": { callback: rotateSidebar(1) },
+		"ctrl+shift+tab": { callback: rotateSidebar(-1) },
+		"shift+pageup": { callback: focusOnApp },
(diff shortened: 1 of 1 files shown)


Delusoire:
It takes a noticeable amount of time (seconds) to add/remove hundreds of elements directly to/from the DOM. All element.remove() and element.append() should happen only after vimOverlay is removed, and when all the changes to its descendants are made, we then can add it back to the DOM. With this, the whole operation becomes quasi-instant.

ohitstom:
> It takes a noticeable amount of time (seconds) to add/remove hundreds of elements directly to/from the DOM. All element.remove() and element.append() should happen only after vimOverlay is removed, and when all the changes to its descendants are made, we then can add it back to the DOM. With this, the whole operation becomes quasi-instant.

i haven't modified this behavior nor do I plan to

Extensions/keyboardShortcut.js
```diff
@@ -1,236 +1,102 @@
...
+	 * @param {boolean | undefined} staticCondition - A static condition.
+	 * @param {(event: KeyboardEvent) => void} callback - Callback function for the event.
 	 */
-
-	// Ctrl + Tab and Ctrl + Shift + Tab to switch sidebar items
-	registerBind("TAB", true, false, false, rotateSidebarDown);
-	registerBind("TAB", true, true, false, rotateSidebarUp);
-
-	// Ctrl + Q to open Queue page
-	registerBind("Q", true, false, false, clickQueueButton);
```
afonsojramos, line 61: Why did this one not get ported?

Extensions/keyboardShortcut.js
```diff
@@ -1,236 +1,102 @@
...
-
-	// Shift + H and Shift + L to go back and forward page
-	registerBind("H", false, true, false, clickNavigatingBackButton);
-	registerBind("L", false, true, false, clickNavigatingForwardButton);
```
afonsojramos, line 65: Why did this one not get ported?

Extensions/keyboardShortcut.js
```diff
@@ -1,236 +1,102 @@
...
-
-	// G and Shift + G to scroll to top and to bottom
-	registerBind("G", false, false, false, appScrollTop);
-	registerBind("G", false, true, false, appScrollBottom);
-
-	// M to Like/Unlike track
-	registerBind("M", false, false, false, Spicetify.Player.toggleHeart);
-
-	// Forward Slash to open search page
-	registerBind("/", false, false, false, openSearchPage);
```
afonsojramos, line 83: Why did this one not get ported?

afonsojramos (changes requested):
Looks good overall! Just want to check the missing shortcuts

Extensions/keyboardShortcut.js
ohitstom: as per the pr message spotify has these built in now just with different syntax and an actual menu (ctrl +?) to see binds, its pointless having dupes....

afonsojramos: Fair on this one! Although the like in the new Keyboard shortcuts seems a bit too complicated and we could probably keep the "m" one, what do you think?

ohitstom: I dont like the idea of adding keybinds just for the sake of a simpler key combination, however if you want it i can do it just lmk

afonsojramos (approved):

Extensions/keyboardShortcut.js
```diff
@@ -1,236 +1,102 @@
...
+		"shift+pageup": { callback: focusOnApp },
+		"shift+pagedown": { callback: focusOnApp },
+
+		// Scroll actions using 'j' and 'k' keys
+		j: { callback: createScrollCallback(SCROLL_STEP) },
+		k: { callback: createScrollCallback(-SCROLL_STEP) },
+
+		// Scroll to the top ('g') or bottom ('Shift+g') of the page
+		g: { callback: scrollToPosition(0) },
+		"shift+g": { callback: scrollToPosition(1) },
```
kyrie25, line 38: Uniformity with the previous diff, modify the functions related accordingly.
```suggestion
		j: { callback: () => createScrollCallback(SCROLL_STEP) },
		k: { callback: () => createScrollCallback(-SCROLL_STEP) },

		// Scroll to the top ('g') or bottom ('Shift+g') of the page
		g: { callback: () => scrollToPosition(0) },
		"shift+g": { callback: () => scrollToPosition(1) },
```

kyrie25 (changes requested):

Extensions/keyboardShortcut.js
ohitstom: have you tested this works since when I did this it did not.

kyrie25: > have you tested this works since when I did this it did not.

> **modify the functions related accordingly**

The functions you wrote in question returns a **callback function** with the memorized values of the constructor function. Modify them to only execute actions instead of another function in combination with this suggestion.

ohitstom: read the code this is already in place.

kyrie25: image

kyrie25: Refrain from closing suggestions when the discussion is not resolved.

ohitstom: suggestions have been added

kyrie25 (approved):

Extensions/keyboardShortcut.js
BachoSeven: Apparently it's Alt+Shift+q; although I liked many of the previous bindings better, as the required less keypresses. Could be nice to be able to customize the bindings, even for those implemented by Spotify themselves.

ohitstom: If you make an issue, provided others agree im happy to reimplement, however we shouldnt really have duplicate keybindings for the same function.

ohitstom:
> It takes a noticeable amount of time (seconds) to add/remove hundreds of elements directly to/from the DOM. All element.remove() and element.append() should happen only after vimOverlay is removed, and when all the changes to its descendants are made, we then can add it back to the DOM. With this, the whole operation becomes quasi-instant.

hey im looking to implement this in my new pr #2860, its my understanding that you mean instead of keeping the vimOverlay in dom and appending / removing hundreds of keys, we instead append all these children **before** appending to document if that makes sense?

image
image

PR #2508, opened 2023-08-11
fix(lyrics-plus/netease): use PyNCMd API

pan93412:
The old xianqiao.wang API seems like removed; therefore, I switch it to PyNCMd, which is actively maintained.

Besides, in Spotify Taiwan, the album name is in Traditional Chinese, which is not aligned with what NCM chooses. Therefore, I have written a simple converter that converts the Traditional Chinese album name to Simplified Chinese so that we can compare the correct album name in such cases.

Note that:

* The PyNCM I choose is from [UnblockNeteaseMusic/music](https://github.com/UnblockNeteaseMusic/server/blob/62b958d83f91ae7b196b9c52f9da870426665df4/src/provider/pyncmd.js#L7). It has been online for at least 3 weeks.
* The PyNCM instance I used have not supported karaoke lyrics, but [PyNCM indeed support fetching it](https://github.com/mos9527/pyncm/blob/master/pyncm/apis/track.py#L136). I may switch or host a PyNCM instance with the latest PyNCM code later.

### DIFF
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -1,26 +1,75 @@
-const ProviderNetease = (function () {
-	const requestHeader = {
-		"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:93.0) Gecko/20100101 Firefox/93.0"
-	};
+/**
+ * @typedef {{
+ *   result: {
+ *     songs: {
+ *       name: string,
+ *       id: number,
+ *		 dt: number,  // duration in ms
+ *       al: {        // album
+ * 			name: string,
+ *       },
+ *     }[],
+ *   },
+ * }} SearchResponse
+ *
+ * @typedef {{
+ * 	title: string,
+ * 	artist: string,
+ * 	album: string,
+ * 	duration: number,
+ * }} Info
+ */
 
-	async function findLyrics(info) {
-		const searchURL = `https://music.xianqiao.wang/neteaseapiv2/search?limit=10&type=1&keywords=`;
-		const lyricURL = `https://music.xianqiao.wang/neteaseapiv2/lyric?id=`;
+const ProviderNetease = (function () {
+	/**
+	 * Search with PyNCM api.
+	 *
+	 * @param {Info} info
+	 * @throw "Cannot find track"
+	 */
+	async function search(info) {
+		const searchURL = `https://pyncmd.apis.imouto.in/api/pyncm?module=cloudsearch&method=GetSearchResult&keyword=`;
 
 		const cleanTitle = Utils.removeExtraInfo(Utils.removeSongFeat(Utils.normalize(info.title)));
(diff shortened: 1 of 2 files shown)


CustomApps/lyrics-plus/ProviderNetease.js
```diff
@@ -1,26 +1,77 @@
...
-		const lyricURL = `https://music.xianqiao.wang/neteaseapiv2/lyric?id=`;
+const ProviderNetease = (function () {
+	/**
+	 * Search with PyNCM api.
+	 *
+	 * @param {Info} info
+	 * @throw "Cannot find track"
+	 */
+	async function search(info) {
+		const searchURL = `https://pyncmd.apis.imouto.in/api/pyncm?module=cloudsearch&method=GetSearchResult&keyword=`;
```
rxri, line 31: From what I have read on the commit you linked, it's a reverse proxy, is there some reason why we can't use `https://pyncmd.vercel.app/api/pyncm` instead?

rxri (changes requested):

CustomApps/lyrics-plus/ProviderNetease.js
pan93412: According to UnblockNeteaseMusic's commit history, [Vercel has been blocked in mainland China](https://github.com/UnblockNeteaseMusic/server/commit/4fe40100f89f7547f2eff33ac84adbd3a8afe5a3). Therefore, UNM's developers are attempting to work around this issue by using a reverse proxy that is accessible in mainland China.

Many people who prefer the lyrics from Netease Cloud Music are located in mainland China. It would be beneficial to use a mirror where they can connect.

rxri: Is there maybe some reverse proxy based on Cloudflare Workers instead or something? Since I think it will be more reliable in the long term, but if there is none, this reverse proxy should be okay then

pan93412: > Is there maybe some reverse proxy based on Cloudflare Workers (since Cloudflare shouldn't be blocked in China iirc) instead or something? Since I think it will be more reliable in the long term, but if there is none, this reverse proxy should be okay then

I may switch to a more reliable PyNCM instance which provides the API to fetch karaoke lyrics while still accessible from mainland China. However, I have not found one. 😢

I will send a PR to both UnblockNeteaseMusic and here when I found one :D

> * The PyNCM instance I used have not supported karaoke lyrics, but [PyNCM indeed support fetching it](https://github.com/mos9527/pyncm/blob/master/pyncm/apis/track.py#L136). I may switch to host a PyNCM instance with the latest PyNCM code later.

rxri: alright 👍

rxri (approved):

kyrie25 (approved):

Lseoksee:
I'm sorry to mention pull that's already finished, but I think it would be better to apply that patch to popupLyrics as well.

PR #2509, opened 2023-08-11
feat(lyrics-plus/netease): normalize album name before comparing

pan93412:
Based on: #2508 (fix(lyrics-plus/netease): use PyNCMd API)


kyrie25 (changes requested):
This has the same diff as #2508 + more, summit only one PR with the relevant content or merge them together.

pan93412:
> This has the same diff as #2508 + more, summit only one PR with the relevant content or merge them together.

Thanks! I have merged this PR to #2508, and I'll close the PR here :D

PR #2513, opened 2023-08-12
feat(wrapper): add webpack related definitions to own class

ohitstom:
image

This PR aims to make it easier for devs to interact with webpack related functions and object, without having to define 5 lines of variables in devtools every reload just to troubleshoot it.

This adds no performance disadvantage as we already define these functions.

If this is closed without proper discussion with the rest of the org i will be reporting whoever closes it to @afonsojramos, it is not a one person decision to deny a PR.

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -292,7 +292,8 @@ window.Spicetify = {
 	},
 	ReactComponent: {},
 	ReactHook: {},
-	ReactFlipToolkit: {}
+	ReactFlipToolkit: {},
+	Webpack: {}
 };
 
 (async function hotloadWebpackModules() {
@@ -334,6 +335,12 @@ window.Spicetify = {
 		.filter(Boolean);
 
 	Object.assign(Spicetify, {
+		Webpack: {
+			chunks: webpackChunkopen,
+			require,
+			modules,
+			functionModules
+		},
 		React: cache.find(m => m?.useMemo),
 		ReactDOM: cache.find(m => m?.createPortal),
 		ReactDOMServer: cache.find(m => m?.renderToString),


rxri (dismissed):
.

ohitstom:
@afonsojramos @itsmeow @theRealPadster @CharlieS1103 @kyrie25 

I would like all of your opinions on this, we should hold a vote since @rxri thinks every one of her decisions are blessed by god whilst not giving a singular reason for them.

itsmeow:
> blocking this pr atm

Why? This PR doesn't even change any behavior, its dev QoL

Aimarekin:
This will be a very convenient change - as an extension developer, Spicetify doesn't always expose what's necessary, or it's not yet released, or it's a marginal feature that won't get implemented into Spicetify. It is sometimes needed to expose something from webpack yourself.

Having devs process webpackChunkOpen on each extension that utilizes it adds unnecessary boilerplate to extensions and extra overhead. It will be more useful and convenient to have this exposed.

This will also simplify Spicetify's development itself - when testing exposing something, pasting the boilerplate into the console every refresh will no longer be necessary. webpackChunkOpen will also now be documented (after type defs are done).

Essentially, the process behind manually extracting something yourself remains the same, but in a more convenient manner and with Spicetify's prefiltering included.

Overall I see no real reason why this should not be implemented - it's a QoL update that will allow devs to develop quicker and more efficiently. Only drawback is if placed under the window these objects will not be garbage collected. This is countered by the fact that each extension that uses webpack will no longer be doing a re-definition and filtering of the chunks.

theRealPadster (approved):
I don't really know what this allows you to do, but I don't see why not if it helps people developing stuff.

Aimarekin:
Currently:
```js
let cachedWebpack = null;

function getWebpack() {
    if (cachedWebpack === null) {
        // Taken from wrapper
        // Force all webpack modules to load
        const require = webpackChunkopen.push([[Symbol()], {}, re => re]);
        const cache = Object.keys(require.m).map(id => require(id));
        const modules = cache
            .filter(module => typeof module === "object")
            .map(module => {
                try {
                    return Object.values(module);
                } catch {}
            })
            .flat();
        const functionModules = modules.filter(module => typeof module === "function");

        cachedWebpack = {
            webpackChunkopen: webpackChunkopen,
            modules: modules,
            cache: cache,
            functionModules: functionModules,
        }
    }
    return cachedWebpack;
}

getWebpack().functionModules.find(m => m.toString().includes("example"));
```

With this update, this would be simplified to
```js
Spicetify.Webpack.functionModules.find(m => m.toString().includes("example"));
```

Aimarekin:
To address concerns raised in the Discord:

- **This change places the burden of responsibility on the developer**: It already is. Spicetify has not exposed something necessary for their extension, and they're forced to expose it themselves. This PR doesn't change that.
- **This change adds instability/sets a new expectation for something that was previously only managed internally**: True, but I don't believe these internals will change in any foreseeable future. If new filters for webpack are added, they should also be exposed here. Only situation where I believe they could change or be removed is if `webpackChunkOpen` itself was modified, in which any case extensions which manually extract from Spotify would break anyway.
- **This makes an unrecommended practice more available to the developer/This should remain intentionally obscure**: Whenever this practice is done it's because it's absolutely necessary and unavoidable. Right now only an experienced developer who knows what they're doing would manually extract, and I don't believe this PR would change that. The practice remains obscure.
A new developer wouldn't initially find anything of use in `Spicetify.Webpack`, it just looks like an advanced internal not worth messing with. Anyone who wants to use it will read the docs and see a big warning saying this should only be used as a last resort. I believe this is enough, no developer would willingly want to manually extract something that's already available. It's worth more investigating something than going out of your way to expose it.
Either way, exposing and documenting this will make developers aware of this necessary last resort instead of leaving them in the dark, navigating through Spicetify's files to figure out how to manually expose things. A developer might also initially believe exposing could only be done in a patch. This would hint them towards this better practice.
Yet another plus is documenting this will make developers more aware of how Spicetify exposes internally, which could help make more contributors for Spicetify. A developer who has been forced to manually expose something will likely contribute to Spicetify exposing it.
- **The only way for developers to know this is bad practice is to make it obscure by forcing them to boilerplate it/make it inconvenient**: This is not true. Manually extracting is already complicated and inconvenient as it is. This PR won't change that. Documentation will be more than enough to warn developers about how `Spicetify.Webpack` is only a last resort.
- **Manually extracting is unstable**: Already was. Manually extracting remains a dangerous practice. As explained previously this will only foment the practice when it is a last resort.

Overall the reality of this PR is that it will only be used by experienced developers and make this process a little easier for them. I can't really imagine any new developer wanting to mess with webpack and manually extracting things they shouldn't or doing it recklessly.

Aimarekin:
If the problem with this PR is documenting and making webpack available, it can be exposed through a window global instead. This will make it completely undocumented.

PR #2520, opened 2023-08-16
fix(wrapper): repair panel functionality broken in .18

ohitstom:
Notes:
- PanelAPI no longer exists
- Panels no longer have number ids instead just using prefixed naming with underscores

Todo:
- Replace old api functions
- Backwards compatibility

Done:
- Repair hooks
- Repair components


rxri:
closing due to 1 month inactivity

PR #2528, opened 2023-08-28
feat: replace `--check-update` with `--update` flag

adufr:
Hello there! 👋

After discussing with @rxri on Discord, this is the follow-up PR of #2527.

This PR adds two flags:
- `--update=check`: to check if an update is available
- `--update=upgrade`: to trigger an upgrade (if available)

It also removes the `--check-update` flag that was a bit misleading because it would trigger an upgrade if one was available.

This solves the issue of not having a proper command to simply check if an update is available and exit after doing so.

I'm open to changes and/or improvements if you have any!

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -263,7 +263,7 @@ func main() {
 	}
 
 	utils.PrintBold("spicetify v" + version)
-	cmd.CheckUpgrade(version)
+	cmd.CheckUpgrade(version, "warning")
 
 	// Chainable commands
 	for _, v := range commands {
--- a/spicetify.go
+++ b/spicetify.go
@@ -430,6 +430,9 @@ upgrade             Upgrade spicetify latest version
 
 -v, --version       Print version number and quit
 
+--update=check      Check for new version
+--update=upgrade    Upgrade to latest available version
+
 When using the "watch" command, any combination of the style (-s), extension (-e), and app (-a) flags can be used (ex. "watch -s -e" or "watch -e -a").
 For config information, run "spicetify -h config".
 For more information and bug report: https://github.com/spicetify/spicetify-cli/`)
--- a/spicetify.go
+++ b/spicetify.go
@@ -6,12 +6,14 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 	"sync"
 
 	colorable "github.com/mattn/go-colorable"
 	"github.com/spicetify/spicetify-cli/src/cmd"
+	spotifystatus "github.com/spicetify/spicetify-cli/src/status/spotify"
 	"github.com/spicetify/spicetify-cli/src/utils"
 )
 
@@ -112,11 +114,30 @@ func init() {
 			switch flagOption {
 			case "check":
 				cmd.CheckUpgrade(version, "info")
+				os.Exit(0)
 			case "upgrade":
-				cmd.Upgrade(version)
+				upgradeStatus := cmd.Upgrade(version)
+				if upgradeStatus {
+					ex, err := os.Executable()
+					if err != nil {
+						ex = "spicetify"
+					}
+
+					spotStat := spotifystatus.Get(utils.FindAppPath())
+					cmds := []string{"backup", "apply"}
+					if !spotStat.IsBackupable() {
+						cmds = append([]string{"restore"}, cmds...)
+					}
+
+					cmd := exec.Command(ex, cmds...)
+					utils.CmdScanner(cmd)
+
+					cmd = exec.Command(ex, strings.Join(commands[:], " "))
+					utils.CmdScanner(cmd)
+
+					os.Exit(0)
+				}
 			}
-			os.Exit(0)
-			return
 		}
 	}
 
@@ -430,8 +451,8 @@ upgrade             Upgrade spicetify latest version
 
 -v, --version       Print version number and quit
 
---update=check      Check for new version
---update=upgrade    Upgrade to latest available version
+--update=check      Check for new version and quit
+--update=upgrade    Upgrade to latest available version (can be used with other commands)
 
 When using the "watch" command, any combination of the style (-s), extension (-e), and app (-a) flags can be used (ex. "watch -s -e" or "watch -e -a").
 For config information, run "spicetify -h config".
--- a/spicetify.go
+++ b/spicetify.go
@@ -62,6 +62,9 @@ func init() {
 		}
 	}
 
+	utils.MigrateConfigFolder()
+	cmd.InitConfig(quiet)
+
 	for _, v := range flags {
 		switch v {
 		case "-c", "--config":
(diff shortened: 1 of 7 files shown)


spicetify.go
```diff
@@ -55,13 +54,22 @@ func init() {
...
 				}
 			} else {
-				flags = append(flags, v)
+				if strings.Contains(v, "=") {
+					split := strings.Split(v, "=")
+					flagOption = split[1]
+					flags = append(flags, split[0])
+				} else {
+					flags = append(flags, v)
+				}
```
adufr, line 63: This could/should later be improved as it only works when we having one single flag

A solution could be to have an array of something like maps, where the key would be the arg name (such as `--update` for eg), and the value would be the arg value (`check`)
I did not include that in this PR as it wouldn't be used for now

spicetify.go
```diff
@@ -438,6 +430,9 @@ upgrade             Upgrade spicetify latest version
 
 -v, --version       Print version number and quit
 
+--update=check      Check for new version
+--update=upgrade    Upgrade to latest available version
```
adufr, line 434: I thought about having a oneliner (something like `--update=<check|upgrade>` but then the flag was too long and it looked bad

So I deliberately regrouped both flags without a new empty line as it looked better IMO

src/cmd/cmd.go
```diff
@@ -299,7 +299,15 @@ func CheckUpgrade(version string) {
...
+		newVersionMsg := "New version available: v" + latestTag + " (current is: v" + version + ")"
+		howToUpgradeMsg := `Run "spicetify upgrade" or using package manager to upgrade spicetify`
+
+		if printLevel == "info" {
+			utils.PrintInfo(newVersionMsg)
+			utils.PrintInfo(howToUpgradeMsg)
+		} else if printLevel == "warning" {
+			utils.PrintWarning(newVersionMsg)
+			utils.PrintWarning(howToUpgradeMsg)
+		}
```
adufr, line 311: Makes more sense to have different printLevels depending on the context
(it shouldn't be "warning" when running `spicetify --update=check`

afonsojramos:
This PR seems to be a mix of over engineering mixed in with personal preferences that don't really match the expected user behaviour.

If we already have `upgrade`, it doesn't really make sense that the `--check-update` upgrades as well, we can agree on that. But having `--update=check` and `--update=check` to override the default behaviour doesn't make any sense at all.

Yes, our situation is a bit more complex than others' because `upgrade`!=`update`, so what I would suggest would be to:

1. Simply rename `--check-update` to `--check-upgrade`.
2. Remove the actual upgrade functionality of the `--check-upgrade`. 

Because, after all, if you want to check if there is an update (/upgrade lol) instead of simply upgrading right away, you probably don't want to upgrade when checking. Probably worth discarding this one and starting from scratch, but I'll let you be the judge of that.

rxri:
discussing it on discord with @afonsojramos

adufr:
I agree with @afonsojramos
Please keep me updated here about the result of that discussion, so that I know what to do 👍

afonsojramos:
Talking with @rxri I now better understand the rationale, but still, neither the PR, nor the current implementation make sense to me. 

If we want `upgrade` to be chainable that's alright, we can shift the current logic for `--check-update` to the upgrade command. And, to be honest, let's simply kill the check optional and simply always check for a version (without upgrading). Why would we not? What do you think @adufr @rxri?

rxri:
The `--check-update` was supposed to be automatic way of upgrading, (restoring), backing up and applying. I don't think making this in `upgrade` command will be okay because from my point of view it should be manual, but if it would be better for user, it's fine

afonsojramos:
> I don't think making this in upgrade command will be okay because from my point of view it should be manual

@rxri in what way is it not manual still? They still need to include the `upgrade` command within the chain.

rxri:
I mean manual in the way of automatic restoration, backup and apply process. Upgrade command doesn't do that and prompts the user to do it. I believe that cli shouldn't handle it by the end-user by default because some stuff can happen

afonsojramos:
> I mean manual in the way of automatic restoration, backup and apply process. Upgrade command doesn't do that and prompts the user to do it. I believe that cli shouldn't handle it by the end-user by default because some issues can occur

If we do it exactly like the current `--check-update` is doing what issues can occur?

rxri:
``--check-update` is sort of a flag that is for advanced users which allows them to do it. It's not a default behavior for everyone

afonsojramos:
@rxri @adufr In that case I see this PR as unneeded and adding unnecessary complexity. And what I would suggest is to rename it to `--check-upgrade`, and to also document in the the help command, as it is not there yet.

adufr:
@afonsojramos ok so I can rename `--check-update` to `--check-upgrade`, add it to the help menu.
But my first motivation, was to have a command that checks if a new version is available without actually upgrading, so that would not address my issue.
Maybe I can just add that to the already existing `-v` flag?

afonsojramos:
@adufr right, thanks for bringing the focus back. 
> There can be issues with backup, restore etc, which are quite frequent from what I've seen on our discord server

But then again, making the upgrade chainable does not affect in any way these issues. They would still have this issues nonetheless. So I'm still pro making `upgrade` chainable.

> Maybe I can just add that to the already existing -v flag?

Yeah, let's do that!

rxri:
`-v` should just return the version of the program like any other cli does. Upgrade command can be chainable sure, but again I was talking about merging the `--check-update` behavior being merged to upgrade cmd

adufr:
> `-v` should just return the version of the program like any other cli does. Upgrade can be chainable sure, but again I was talking about merging the `--check-update` behavior to upgrade cmd. I have nothing against the chainable cmd

I don't understand you, a week ago you were ready to make it so that every command would run the "checking for update" function, and now you're against checking for new update in the `-v` flag..?

I think I'm losing my time here, I made multiple suggestions, which many of them were in my opinion good solutions, and everytime you're going against it without actually proposing a solution.

I need to remember you that my very first proposition was to add a new command to check for updates (#2527), and I still think that it was a good idea because it would not have changed anything about the way the cli works today...

rxri:
That every command will execute it? No, I didn't say it. We didn't even talk about that. So you might've misunderstood me. At this point I'm okay with just doing command "check-update" or something similar and renaming the "--check-update" flag to "--check-upgrade" or maybe better name like "--exec-upgrade"

rxri:
That every command would execute it? I didn’t comment on it so I had and still have no take on it. You might've misunderstood me. At this point I'm okay with the first implementation and renaming the "--check-update" flag to "--check-upgrade" or maybe better name like "--exec-upgrade", but this is up to @afonsojramos at this point

adufr:
@afonsojramos after discussing with @rxri on discord again, here's our take:

1. if you agree with it, I can reopen #2527, which would be an easy solution for having a command that actually checks for an update without upgrading

2. we can rename the existing `--check-update` flag, which is a bit misleading (and will be even more if there's a `check-update` command).
I don't really have any opinion about it's new name so I'm open for suggestions, but I will also add it to the help section (because it's currently not documented)

rxri:
so... @afonsojramos what are we going to do with this because this pull request is kinda opened for a long time now

afonsojramos:
Yeah, not with this PR. Let's close this for now. Thanks for the PR though @adufr

PR #2537, opened 2023-08-31
fix: adapt extensions to new player data structure

rxri:


.github/workflows/prettier.yml
```diff
@@ -16,6 +16,6 @@ jobs:
       - name: Prettify code
         uses: creyD/prettier_action@v4.3
         with:
-          prettier_options: --write .
+          prettier_options: --write . --config .prettierrc
```
afonsojramos, line 19: Pretty sure that prettier tries to find the config file, so no need to specify it
image
```suggestion
          prettier_options: --write .
```

afonsojramos (approved):

.github/workflows/prettier.yml
afonsojramos: The action runs the prettier CLI, therefore you can assume that it does. But feel free to test by changing a parameter of the config ;)

PR #2543, opened 2023-09-01
fix(shuffle+): add context support for Queue

Delusoire:
PlayerAPI.addToQueue really only takes an array of uri objects, the rest is ignored inside PlayerAPI._queue.createQueueItem
To set the playing context, we can just use PlayerAPI.updateContext

### DIFF
--- a/Extensions/shuffle+.js
+++ b/Extensions/shuffle+.js
@@ -463,7 +463,29 @@
 
 		await Spicetify.Platform.PlayerAPI.clearQueue();
 
-		Spicetify.Platform.PlayerAPI.addToQueue(list.map(uri => ({ uri })));
+		{
+			const { _queue, _client } = Spicetify.Platform.PlayerAPI._queue;
+			const { prevTracks, queueRevision } = _queue;
+
+			const nextTracks = list.map(uri => ({
+				contextTrack: {
+					uri,
+					uid: "",
+					metadata: {
+						is_queued: "false"
+					}
+				},
+				removed: [],
+				blocked: [],
+				provider: "context"
+			}));
+
+			_client.setQueue({
+				nextTracks,
+				prevTracks,
+				queueRevision
+			});
+		}
 
 		if (context) {
 			const { sessionId } = Spicetify.Platform.PlayerAPI.getState();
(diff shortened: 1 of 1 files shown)


Extensions/shuffle+.js
```diff
@@ -463,15 +463,34 @@
...
-		);
+		{
+			const { _queue, _client } = Spicetify.Platform.PlayerAPI._queue;
+			const { prevTracks, queueRevision } = _queue;
+
+			const nextTracks = list.map(uri => ({
+				contextTrack: {
+					uri,
+					uid: "",
+					metadata: {
```
rxri, line 474: if there's no uid, it doesn't need to be set

Extensions/shuffle+.js
```diff
@@ -463,15 +463,34 @@
...
+						is_queued: "false"
+					}
+				},
+				removed: [],
+				blocked: [],
```
rxri, line 479: no need to set it either

rxri (changes requested):

Extensions/shuffle+.js
Delusoire: Actually, it has to be there, The internal Spotify methods that work on this data don't null check, instead, when they call a function they use a fromPartial that fills the rest of the object with default values. Without the default values, you are guaranteed to run into errors.

Extensions/shuffle+.js
Delusoire: Same here

rxri (approved):

Extensions/shuffle+.js
```diff
@@ -459,19 +459,39 @@
 	async function Queue(list, context = null, type) {
 		const count = list.length;
 
+		// Delimits the end of our list, as Spotify may add new context tracks to the queue
 		list.push("spotify:delimiter");
 
-		await Spicetify.Platform.PlayerAPI.clearQueue();
+		{
```
rxri, line 465: Btw why do you execute it in separate block? There's no variable that may conflict with other ones I believe

rxri:
also run prettier on the file

Extensions/shuffle+.js
Delusoire: Visual clarity, The function is getting pretty long and I just figured that It would be nice to delimit that block

rxri: since it's not needed, please remove it

PR #2547, opened 2023-09-02
fix(wrapper): playbar.button match Spotify structure

ohitstom:
Spotify has started wrapping playbar buttons in a span instead of displaying as block, to prevent weird padding behaviour between ours and their buttons; we should do the same.

before:
image

after:
image

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -1842,7 +1842,9 @@ Spicetify.Playbar = (function () {
 		constructor(label, icon, onClick = () => {}, disabled = false, active = false, registerOnCreate = true) {
 			this.element = document.createElement("button");
 			this.element.classList.add("main-genericButton-button");
-			this.element.style.display = "block";
+			this.iconElement = document.createElement("span");
+			this.iconElement.classList.add("IconWrapper__Wrapper-sc-16usrgb-0", "Wrapper-sm-only");
+			this.element.appendChild(this.iconElement);
 			this.icon = icon;
 			this.onClick = onClick;
 			this.disabled = disabled;
@@ -1871,7 +1873,7 @@ Spicetify.Playbar = (function () {
 				input = `<svg height="16" width="16" viewBox="0 0 16 16" fill="currentColor" stroke="currentColor">${Spicetify.SVGIcons[input]}</svg>`;
 			}
 			this._icon = input;
-			this.element.innerHTML = input;
+			this.iconElement.innerHTML = input;
 		}
 		get onClick() {
 			return this._onClick;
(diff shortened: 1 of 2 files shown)


jsHelper/spicetifyWrapper.js
```diff
@@ -1842,7 +1842,9 @@ Spicetify.Playbar = (function () {
 		constructor(label, icon, onClick = () => {}, disabled = false, active = false, registerOnCreate = true) {
 			this.element = document.createElement("button");
 			this.element.classList.add("main-genericButton-button");
-			this.element.style.display = "block";
+			this.iconElement = document.createElement("span");
+			this.iconElement.classList.add("IconWrapper__Wrapper-sc-16usrgb-0", "Wrapper-sm-only");
```
afonsojramos, line 1846: I don't see this as being a valuable addition as we'd be using Spotify's weird classnames, which we already try to run off of using the css mapper. 🤔

ohitstom: i literally explained the reason in the description.

Also we do this constantly and it's unchanged since we have a css map.

ohitstom: maybe it's achievable using `Spicetify.ReactComponent.IconComponent` like we do with the navbar icon links.

ohitstom:
@theRealPadster could you take a look and review this pr if you get a chance please!

theRealPadster:
> @theRealPadster could you take a look and review this pr if you get a chance please!

I'm not sure I really understand what's going on? My Spotify doesn't seem to have any spans, though maybe it's outdated? 

image

ohitstom:
> > @theRealPadster could you take a look and review this pr if you get a chance please!
> 
> I'm not sure I really understand what's going on? My Spotify doesn't seem to have any spans, though maybe it's outdated? 
> 
> image
> 

You expanded the wrong div, that's the device picker subtree 👍

theRealPadster:
> You expanded the wrong div, that's the device picker subtree 👍
> 
> for reference this is being tested on 2.20 which has the breaking changes for lyrics plus, however the spans are visible on 2.19 too

Ah, I'll have to check my version and probably let it update and look again tomorrow or whenever I get a chance. Will do though.

ohitstom:
> > You expanded the wrong div, that's the device picker subtree 👍
> > 
> > for reference this is being tested on 2.20 which has the breaking changes for lyrics plus, however the spans are visible on 2.19 too
> 
> Ah, I'll have to check my version and probably let it update and look again tomorrow or whenever I get a chance. Will do though. 
> 
> 

thanks!

theRealPadster (approved):
Ah yeah I see it now. Updated to x.x.20

afonsojramos (approved):
Looking good! Thanks @ohitstom, and sorry for the delays!

PR #2565, opened 2023-09-19
fix(lyrics-plus): synced line width

SunsetTechuila:
removes unnecessary horizontal scrollbar for me when window width is more than 2500px (98% of my display width)

before:

Без имени-1

after:

Без имени-2


SunsetTechuila:
breaks lyrics right/center alignment

PR #2573, opened 2023-09-27
fix(wrapper): support for `UserAPI._product_state_service`

SunsetTechuila:


rxri (approved):

jsHelper/spicetifyWrapper.js
```diff
@@ -1025,34 +1025,37 @@ Spicetify.SVGIcons = {
...
-	if (!Spicetify.Platform?.UserAPI?._product_state) {
+	if (!Spicetify.Platform?.UserAPI) {
 		setTimeout(waitUserAPI, 1000);
 		return;
 	}
 
 	let subRequest;
 
+	// product_state was renamed to product_state_service in Spotify 1.2.21
+	const productState = Spicetify.Platform.UserAPI._product_state || Spicetify.Platform.UserAPI._product_state_service;
```
rxri, line 1036: actually, you should leave the optional chaining
```suggestion
	const productState = Spicetify.Platform.UserAPI?._product_state || Spicetify.Platform.UserAPI?._product_state_service;
```

rxri (changes requested):

jsHelper/spicetifyWrapper.js
SunsetTechuila: This line will not be executed if UserAPI is undefined, will it?

rxri: Huh. JavaScript can simply throw error because it will try to access variable that doesn't exist in this object - it's better to be safe

rxri (approved):

afonsojramos (approved):

PR #2599, opened 2023-10-12
feat(lyrics-plus): add cutlet japanese transliteration method

faishalirwn:
The current japanese romaji method (kuroshiro) produces a lot of incorrect transliterations. I have compiled some of my findings here: https://www.notion.so/faishalirwn/e9c8db6a52fb43b0acaa6b89e600039a?v=638e78a3c0a540a1ac93bcede6915bad

There is a better library for Japanese transliteration called cutlet: https://github.com/polm/cutlet and it's actively maintained unlike kuroshiro. Unfortunately, it is written in Python, so I have created my own API for it: https://cutlet.1sal.me/docs. The source code for the API is available here: https://github.com/faishalirwn/cutlet-api

This PR adds an alternative romaji method that uses the cutlet API. The API can also correct wrong romaji using the `correction.json` file in the source code repo https://github.com/faishalirwn/cutlet-api/blob/main/corrections.json.

### DIFF
--- a/CustomApps/lyrics-plus/OptionsMenu.js
+++ b/CustomApps/lyrics-plus/OptionsMenu.js
@@ -121,6 +121,7 @@ const TranslationMenu = react.memo(({ friendlyLanguage, hasTranslation }) => {
 				modeOptions = {
 					furigana: "Furigana",
 					romaji: "Romaji",
+					romajiCutlet: "Romaji (Cutlet)",
 					hiragana: "Hiragana",
 					katakana: "Katakana"
 				};
--- a/CustomApps/lyrics-plus/Translator.js
+++ b/CustomApps/lyrics-plus/Translator.js
@@ -98,11 +98,25 @@ class Translator {
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


rxri:
i'm really against this because it's the one who hosts the api and is not used widely and it can go down at any time

surfbryce:
> i'm really against this because you're the one who hosts the api and is not used widely and it can go down at any time. I understand that there are issues with kuroshiro (as seen in your findings) but this one doesn't need self-hosted API to work so atm we will stick with this one

I also want to add that I looked through the source of Cutlet and it would be far better just to convert this to Javascript and have this run locally than through a web-API. There’s no OS level dependencies so if OP really wants it would be better just to start converting it. It would also be contributing to a wider audience more then just Lyrics Plus

rxri:
If they can port it to JavaScript module I will be more than happy to add it

afonsojramos:
Agree to what has been said above!

faishalirwn:
Thank you, I understand the concern, I'll try to learn to the source code and see if I can port it to JavaScript

PR #2601, opened 2023-10-13
feat(preprocess): implement snackbar api

### LINKED ISSUE #2600: Expose new notification types

Spotify now has new notifications it seems to use on 99% of stuff, it would be great if devs can access these.
this is their html structure:
image

this is how they look:
image

### ➕ Additional Information

These new notifications stay forever when hovered over and have intractability so if you click them they disappear and probably a lot more internal functionality we dont yet know of...

I found very little in xpui with a quick look looks to be a lot more of the relevant keywords within vendor

https://mui.com/material-ui/api/snackbar/

ohitstom:
implements Spotify's new (1.2.22.982) notification system resolves #2600
image

### "info" / iconVariation
```js
Spicetify.Snackbar.enqueueSnackbar('I am an exception', {
  autoHideDuration: 3000,
    style: {
        backgroundColor: "purple"
    },
    onClose() {
        console.log('closed')
    },
    variant: 'info',
})
```

image

#### HTML
```js
Spicetify.Snackbar.enqueueSnackbar(
  `<div class="TcJ5lUfQ25AJNoCrWO5Z">
	  <div class="xaCo5LWfBMxqfWjUA3o2">
		  <img src="spotify:image:ab67706c0000da849e94705a00ff5f32789c8b8e" alt="" data-encore-id="image" class="Image-sc-1u215sg-3 Image-image">
	  </div>  
	   <div class="r4Hbxvv02KfOVeZ_v335">
		   <span data-encore-id="type" class="Type__TypeElement-sc-goli3j-0 TypeElement-ballad-textBase-type">
			   Added to 
			   <span data-encore-id="type" class="Type__TypeElement-sc-goli3j-0 TypeElement-balladBold-textBase-type">
					▸ Tasteful 👌
			   </span>
			</span>
		</div>
	</div>`,
  { persist: true },
);
```
image

#### Plaintext
```js
Spicetify.Snackbar.enqueueSnackbar('barebones snack')
```

image

### "headless" / enqueueCustomSnackbar
```js
Spicetify.Snackbar.enqueueSnackbar({
  key: "test",
  variant: "headless",
  children: Spicetify.ReactComponent.Slider(),
  style: { backgroundColor: "red" },
});

Spicetify.Snackbar.enqueueCustomSnackbar(
  { key: "test", children: Spicetify.ReactComponent.Slider() },
  { style: { backgroundColor: "red" } },
);
```

image

### enqueueImageSnackbar
```js
Spicetify.Snackbar.enqueueImageSnackbar({
  message: "test",
  imageSrc:
    "https://images-ext-1.discordapp.net/external/i1dw80RDh1sbEY_rfZNwpY_4HpuPSDdoHj8NL9zNPZE/https/artworks.thetvdb.com/banners/v4/series/423030/backgrounds/64fe06fb91fee.jpg?width=921&height=518",
});
```
image

## React Components
```js
Spicetify.Snackbar.enqueueCustomSnackbar({
  key: "test",
  children: Spicetify.ReactComponent.Snackbar.wrapper({
    children: Spicetify.ReactComponent.Snackbar.simpleLayout({
      leading: "start",
      center: "center",
      trailing: Spicetify.ReactComponent.Snackbar.ctaText({
        ctaText: "undo",
        onCtaClick: Spicetify.Snackbar.handleCloseSnack("test"),
      }),
    }),
  }),
});

```
image

## Relevant Documents
this is pretty much all users need to get started -> https://notistack.com/api-reference + https://mui.com/material-ui/react-snackbar

enqueue valid props:
```
  message: o,
  open: !0,
  entered: !1,
  requestClose: !1,
  persist: d("persist"),
  action: d("action"),
  content: d("content"),
  variant: d("variant"),
  anchorOrigin: d("anchorOrigin"),
  disableWindowBlurListener: d("disableWindowBlurListener"),
  autoHideDuration: d("autoHideDuration"),
  hideIconVariant: d("hideIconVariant"),
  TransitionComponent: d("TransitionComponent"),
  transitionDuration: d("transitionDuration"),
  TransitionProps: d("TransitionProps", !0),
  iconVariant: d("iconVariant", !0),
  style: d("style", !0),
  SnackbarProps: d("SnackbarProps", !0),
  className: a(n.props.className, c.className)
```

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -406,6 +406,7 @@ window.Spicetify = {
 			ButtonTertiary: modules.find(m => m?.render && m?.displayName === "ButtonTertiary"),
 			Snackbar: {
 				//...Spicetify.Snackbar.props.Components, -> doesnt exist for a while
+				wrapper: functionModules.find(m => m.toString().includes('encore-light-theme')),
 				simpleLayout: functionModules.find(m => m.toString().includes("leading")),
 				ImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar")),
 				ctaText: functionModules.find(m => m.toString().includes("ctaText"))
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -126,6 +126,7 @@ window.Spicetify = {
 			"LocalStorage",
 			"Queue",
 			"removeFromQueue",
+			"notification",
 			"showNotification",
 			"Menu",
 			"ContextMenu",
@@ -445,6 +446,25 @@ window.Spicetify = {
 	if (playlistMenuChunk) Spicetify.ReactComponent.PlaylistMenu = Object.values(require(playlistMenuChunk[0])).find(m => typeof m === "function");
 
 	if (Spicetify.Color) Spicetify.Color.CSSFormat = modules.find(m => m?.RGBA);
+
+	// Combine snackbar and notification
+	(async function bindShowNotification() {
+		if (!(Spicetify.Snackbar && Spicetify.notification)) {
+			setTimeout(bindShowNotification, 10);
+			return;
+		}
+
+		Spicetify.showNotification = (message, isError = false, msTimeout, snack = true) => {
+			if (snack) {
+				Spicetify.Snackbar.enqueueSnackbar(message, {
+					variant: isError ? "error" : "default",
+					autoHideDuration: msTimeout
+				});
+			} else {
+				Spicetify.notification(message, isError, msTimeout);
+			}
+		};
+	})();
 
 	// Image color extractor
 	(async function bindColorExtractor() {
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -402,6 +402,11 @@ window.Spicetify = {
 			ButtonPrimary: modules.find(m => m?.render && m?.displayName === "ButtonPrimary"),
 			ButtonSecondary: modules.find(m => m?.render && m?.displayName === "ButtonSecondary"),
 			ButtonTertiary: modules.find(m => m?.render && m?.displayName === "ButtonTertiary"),
+			Snackbar: {
+				// somehow append Spicetify.Snackbar.props.Components
+				simpleLayout: functionModules.find(m => m.toString().includes("leading")),
+			   	simpleImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar"))
+			},
 			...Object.fromEntries(menus)
 		},
 		ReactHook: {
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -459,7 +459,7 @@ window.Spicetify = {
 				Spicetify.notification(message, isError, msTimeout);
 				return;
 			}
-			
+
 			Spicetify.Snackbar.enqueueSnackbar(message, {
 				variant: isError ? "error" : "default",
 				autoHideDuration: msTimeout
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -405,10 +405,8 @@ window.Spicetify = {
 			ButtonSecondary: modules.find(m => m?.render && m?.displayName === "ButtonSecondary"),
 			ButtonTertiary: modules.find(m => m?.render && m?.displayName === "ButtonTertiary"),
 			Snackbar: {
-				//...Spicetify.Snackbar.props.Components, -> doesnt exist for a while
-				wrapper: functionModules.find(m => m.toString().includes('encore-light-theme')),
+				wrapper: functionModules.find(m => m.toString().includes("encore-light-theme")),
 				simpleLayout: functionModules.find(m => m.toString().includes("leading")),
-				ImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar")),
 				ctaText: functionModules.find(m => m.toString().includes("ctaText"))
 			},
 			...Object.fromEntries(menus)
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -126,7 +126,7 @@ window.Spicetify = {
 			"LocalStorage",
 			"Queue",
 			"removeFromQueue",
-			"notification",
+			"sendNotification",
 			"showNotification",
 			"Menu",
 			"ContextMenu",
@@ -449,20 +449,21 @@ window.Spicetify = {
 
 	// Combine snackbar and notification
 	(async function bindShowNotification() {
-		if (!(Spicetify.Snackbar && Spicetify.notification)) {
+		if (!(Spicetify.Snackbar || Spicetify.sendNotification)) {
 			setTimeout(bindShowNotification, 10);
 			return;
 		}
 
-		Spicetify.showNotification = (message, isError = false, msTimeout, snack = true) => {
-			if (snack) {
-				Spicetify.Snackbar.enqueueSnackbar(message, {
-					variant: isError ? "error" : "default",
-					autoHideDuration: msTimeout
-				});
-			} else {
+		Spicetify.showNotification = (message, isError = false, msTimeout) => {
+			if (!Spicetify.Snackbar?.enqueueSnackbar) {
 				Spicetify.notification(message, isError, msTimeout);
+				return;
 			}
+			
+			Spicetify.Snackbar.enqueueSnackbar(message, {
+				variant: isError ? "error" : "default",
+				autoHideDuration: msTimeout
+			});
 		};
 	})();
 
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -456,7 +456,7 @@ window.Spicetify = {
 
 		Spicetify.showNotification = (message, isError = false, msTimeout) => {
 			if (!Spicetify.Snackbar?.enqueueSnackbar) {
-				Spicetify.notification(message, isError, msTimeout);
+				Spicetify.sendNotification(message, isError, msTimeout);
 				return;
 			}
 
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -163,7 +163,8 @@ window.Spicetify = {
 			"ReactQuery",
 			"Color",
 			"extractColorPreset",
-			"ReactDOMServer"
+			"ReactDOMServer",
+			"Snackbar"
 		];
 
 		const PLAYER_METHOD = [
@@ -225,7 +226,8 @@ window.Spicetify = {
 			"RemoteConfigProvider",
 			"ButtonPrimary",
 			"ButtonSecondary",
-			"ButtonTertiary"
+			"ButtonTertiary",
+			"Snackbar"
 		];
 
 		const REACT_HOOK = ["DragHandler", "usePanelState", "useExtractedColor"];
@@ -403,9 +405,9 @@ window.Spicetify = {
 			ButtonSecondary: modules.find(m => m?.render && m?.displayName === "ButtonSecondary"),
 			ButtonTertiary: modules.find(m => m?.render && m?.displayName === "ButtonTertiary"),
 			Snackbar: {
-				// somehow append Spicetify.Snackbar.props.Components
+				//...Spicetify.Snackbar.props.Components, -> doesnt exist for a while
 				simpleLayout: functionModules.find(m => m.toString().includes("leading")),
-			   	simpleImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar"))
+				ImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar"))
 			},
 			...Object.fromEntries(menus)
 		},
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -407,7 +407,8 @@ window.Spicetify = {
 			Snackbar: {
 				//...Spicetify.Snackbar.props.Components, -> doesnt exist for a while
 				simpleLayout: functionModules.find(m => m.toString().includes("leading")),
-				ImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar"))
+				ImageLayout: functionModules.find(m => m.toString().includes("enqueueCustomSnackbar")),
+				ctaText: functionModules.find(m => m.toString().includes("ctaText"))
 			},
 			...Object.fromEntries(menus)
 		},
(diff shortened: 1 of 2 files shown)


ohitstom:
dont merge yet !!

src/preprocess/preprocess.go
```diff
@@ -354,6 +354,12 @@ Spicetify.React.useEffect(() => {
 		`case [\w$.]+BuddyFeed:(?:return ?|[\w$]+=)[\w$?]*(?:\([\w$.,]+\)\([\w(){},.:]+)?;(?:break;)?(?:case [\w$.]+:(?:return ?|[\w$]+=)[\w$?]*(?:\([\w$.,]+\)\([\w(){},.:]+)?[\w:]*;(?:break;)?)*default:(?:return ?|[\w$]+=)`,
 		`${0} Spicetify.Panel?.render()??`)
 
+	// Snackbar https://mui.com/material-ui/react-snackbar/
+	utils.Replace(
+		&input,
+		`\b\w\s*\(\)\s*[^;,]*enqueueCustomSnackbar:\s*(\w)\s*[^;]*;`,
+		`${0} globalThis.Spicetify.Snackbar.enqueueCustomSnackbar = ${1};`)
```
rxri, line 361: Why you use `globalThis`? Just do
```suggestion
		`${0} Spicetify.Snackbar.enqueueCustomSnackbar = ${1};`)
```

src/preprocess/preprocess.go
```diff
@@ -436,6 +442,12 @@ if (${1}.popper?.firstChild?.id === "context-menu") {
 		`([\w$]+)=((?:function|\()([\w$.,{}()= ]+(?:springConfig|overshootClamping)){2})`,
 		`${1}=Spicetify.ReactFlipToolkit.spring=${2}`)
 
+	// Snackbar https://mui.com/material-ui/react-snackbar/
+	utils.Replace(
+		&input,
+		`\w+\s*=\s*\w\.call\(this,[^)]+\)\s*\|\|\s*this\)\.enqueueSnackbar`,
+		`globalThis.Spicetify.Snackbar = ${0}`)
```
rxri, line 449: same as above
```suggestion
		`Spicetify.Snackbar = ${0}`)
```

rxri (changes requested):
also, since it's only in 1.2.22 I don't know if it's worth it because spicetify supports versions from 1.2.0 to latest one so having it work only one version could potentially not be good for some users if developer would use it and we're not going to drop support for all versions except latest only because of this change

ohitstom:
> also, since it's only in 1.2.22 I don't know if it's worth it because spicetify supports versions from 1.2.0 to latest one so having it work only one version could potentially not be good for some users if developer would use it and we're not going to drop support for all versions except latest only because of this change

I think this is very worth it Spotify will soon not use the old notification system at all, we also have many features like panels that are unsupported (or was).

I don't see how we'd have to drop support, like panels if it doesn't find the regex then it isn't added and the few devs that use this will have to either tell users to update or not use their stuff.

rxri:
> we also have many features like panels that are unsupported

Panels were backported by kyrie to work on every supported version, so I don't really understand why you brought it up. I mean it doesn't work because Spotify made it "somewhat impossible" to inject to. Your situation is that you expect developers now to use two variations of notification libraries and check whichever is available. Majority of extensions or custom apps work on every supported version by spicetify. I did not see the extension that only works on specific version.

> we also have many features like panels that are unsupported

what features are unsupported except panels in latest spicetify release?

ohitstom:
> > we also have many features like panels that are unsupported
> 
> Panels were backported by kyrie to work on every supported version, so I don't really understand why you brought it up. I mean it doesn't work because Spotify made it "somewhat impossible" to inject to. Your situation is that you expect developers now to use two variations of notification libraries and check whichever is available. Majority of extensions or custom apps work on every supported version by spicetify. I did not see the extension that only works on specific version.
> 
> > we also have many features like panels that are unsupported
> 
> what features are unsupported except panels in latest spicetify release?

im not expecting devs to do that at all, its simply an option they can choose to use if they want to match spotify styling.

as for panels theyve been unsupported for many versions now and i bring it up because we are making no moves to fix or even remove it despite them still being 'supported' in docs

i really dont see this as causing a compatibility issue its simply up to devs, im sure the use of many of the react components whos props are constantly changing would also cause compat issues but again thats up for devs to solve.

rxri:
Talked with a few developers about it, and they are okay with adding it so am I then. However, did you try and find snackbar inside webpack modules? Since it would be much better if we would use webpack for it

ohitstom:
> Talked with a few developers about it, and they are okay with adding it so am I then. However, did you try and find snackbar inside webpack modules? Since it would be much better if we would use webpack for it

i tried my hardest, but it was SUPER buried (or im just dumb), maybe you could have a quick look since you are better with webpack? if you check the regex result you can see the rough area

my issue was 99% of the important funcs where in form function e() with the parent func holding the actual props and having 0 identifiable keywords (like the current webpack selection i have for simpleImage react component above which i still need to fix somehow)

ohitstom:
in that case!

other than docs (which this pr can temporarily serve as) and testing that the regex doesnt completely destroy xpui on older versions (it shouldnt since all the keywords i used for the regex arent used elsewhere) this PR is ready for testing and merging 🎊

ohitstom:
```js
function showNotificationFallback(message, isError = false, msTimeout) {
  Spicetify.Snackbar
    ? Spicetify.Snackbar.enqueueSnackbar(message, {
        variant: isError ? "error" : "default",
        autoHideDuration: msTimeout
      })
    : Spicetify.showNotification(message, isError, msTimeout);
}
Spicetify.showNotificationFallback = showNotificationFallback;
```

made some code to showcase how easy it is to convert between the two apis, this could very easily be turnt into an official thing in the wrapper if you wanted it to? we could even convert the shownotification function directly instead to prefer snackbar if its available (with an added override value at the end of course)

rxri:
I wouldn't put it in the another function that developers would need to adopt. I think making `showNotification` use the Snackbar is better. When Snackbar isn't available, it should use the old way (`showNotification`) instead. You can add it to this pull request too

ohitstom:
> Personally, I wouldn't put it in the another function that developers would need to adopt. I think making `showNotification` use the Snackbar is better. When Snackbar isn't available, it should use the old way (`showNotification`) instead. You can add it to this pull request too

sounds good I think that's the best plan of action!

src/preprocess/preprocess.go
```diff
@@ -354,6 +354,17 @@ Spicetify.React.useEffect(() => {
 		`case [\w$.]+BuddyFeed:(?:return ?|[\w$]+=)[\w$?]*(?:\([\w$.,]+\)\([\w(){},.:]+)?;(?:break;)?(?:case [\w$.]+:(?:return ?|[\w$]+=)[\w$?]*(?:\([\w$.,]+\)\([\w(){},.:]+)?[\w:]*;(?:break;)?)*default:(?:return ?|[\w$]+=)`,
 		`${0} Spicetify.Panel?.render()??`)
 
+	// Snackbar https://mui.com/material-ui/react-snackbar/
+	utils.Replace(
+		&input,
+		`\b\w\s*\(\)\s*[^;,]*enqueueCustomSnackbar:\s*(\w)\s*[^;]*;`,
+		`${0} Spicetify.Snackbar.enqueueCustomSnackbar = ${1};`)
```
rxri, line 361:
```suggestion
		`${0}Spicetify.Snackbar.enqueueCustomSnackbar=${1};`)
```

src/preprocess/preprocess.go
```diff
@@ -354,6 +354,17 @@ Spicetify.React.useEffect(() => {
...
+
+	utils.Replace(
+		&input,
+		`\(\({[^}]*,\s*imageSrc`,
+		`Spicetify.Snackbar.enqueueImageSnackbar = ${0}`)
```
rxri, line 366:
```suggestion
		`Spicetify.Snackbar.enqueueImageSnackbar=${0}`)
```

src/preprocess/preprocess.go
```diff
@@ -436,6 +447,12 @@ if (${1}.popper?.firstChild?.id === "context-menu") {
 		`([\w$]+)=((?:function|\()([\w$.,{}()= ]+(?:springConfig|overshootClamping)){2})`,
 		`${1}=Spicetify.ReactFlipToolkit.spring=${2}`)
 
+	// Snackbar https://mui.com/material-ui/react-snackbar/
+	utils.Replace(
+		&input,
+		`\w+\s*=\s*\w\.call\(this,[^)]+\)\s*\|\|\s*this\)\.enqueueSnackbar`,
+		`Spicetify.Snackbar = ${0}`)
```
rxri, line 454:
```suggestion
		`Spicetify.Snackbar=${0}`)
```

jsHelper/spicetifyWrapper.js
```diff
@@ -439,6 +447,25 @@ window.Spicetify = {
...
+		Spicetify.showNotification = (message, isError = false, msTimeout, snack = true) => {
+			if (snack) {
+				Spicetify.Snackbar.enqueueSnackbar(message, {
+					variant: isError ? "error" : "default",
+					autoHideDuration: msTimeout
+				});
+			} else {
+				Spicetify.notification(message, isError, msTimeout);
+			}
+		};
```
rxri, line 466: That's not what I meant
```suggestion
		Spicetify.showNotification = (message, isError = false, msTimeout) => {
			if (!Spicetify.Snackbar?.enqueueSnackbar) {
				Spicetify.notification(message, isError, msTimeout);
				return;
			}
			
			Spicetify.Snackbar.enqueueSnackbar(message, {
				variant: isError ? "error" : "default",
				autoHideDuration: msTimeout
			});
		};
```

src/preprocess/preprocess.go
```diff
@@ -273,7 +273,7 @@ func exposeAPIs_main(input string) string {
 	utils.Replace(
 		&input,
 		`(?:\w+ |,)([\w$]+)=(\([\w$]+=[\w$]+\.dispatch)`,
-		`;globalThis.Spicetify.showNotification=(message,isError=false,msTimeout)=>${1}({message,feedbackType:isError?"ERROR":"NOTICE",msTimeout});const ${1}=${2}`)
+		`;globalThis.Spicetify.notification=(message,isError=false,msTimeout)=>${1}({message,feedbackType:isError?"ERROR":"NOTICE",msTimeout});const ${1}=${2}`)
```
rxri, line 276:
```suggestion
		`;globalThis.Spicetify.sendNotification=(message,isError=false,msTimeout)=>${1}({message,feedbackType:isError?"ERROR":"NOTICE",msTimeout});const ${1}=${2}`)
```

jsHelper/spicetifyWrapper.js
```diff
@@ -126,6 +126,7 @@ window.Spicetify = {
 			"LocalStorage",
 			"Queue",
 			"removeFromQueue",
+			"notification",
 			"showNotification",
```
rxri, line 130:
```suggestion
			"sendNotification",
			"showNotification",
```

jsHelper/spicetifyWrapper.js
```diff
@@ -402,6 +405,11 @@ window.Spicetify = {
 			ButtonPrimary: modules.find(m => m?.render && m?.displayName === "ButtonPrimary"),
 			ButtonSecondary: modules.find(m => m?.render && m?.displayName === "ButtonSecondary"),
 			ButtonTertiary: modules.find(m => m?.render && m?.displayName === "ButtonTertiary"),
+			Snackbar: {
```
rxri, line 408: just to be safe
```suggestion
			Snackbar: {
				...Spicetify.Snackbar,
```

rxri (changes requested):

jsHelper/spicetifyWrapper.js
```diff
@@ -439,6 +447,25 @@ window.Spicetify = {
 
 	if (Spicetify.Color) Spicetify.Color.CSSFormat = modules.find(m => m?.RGBA);
 
+	// Combine snackbar and notification
+	(async function bindShowNotification() {
+		if (!(Spicetify.Snackbar && Spicetify.notification)) {
```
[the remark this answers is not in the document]
ohitstom: but the snack bar definition is entirely xpui and isn't loaded instantly? the other snackbar definition is for react components unrelated

jsHelper/spicetifyWrapper.js
ohitstom: isn't that the same thing just with the inability for devs to override it? guess that works regardless though 👍

jsHelper/spicetifyWrapper.js
ohitstom: why would we want the normal snackbar functions in the react components?

rxri: https://github.com/spicetify/spicetify-cli/pull/2601#discussion_r1362322493

jsHelper/spicetifyWrapper.js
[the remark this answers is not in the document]
rxri: updated suggestion then

jsHelper/spicetifyWrapper.js
ohitstom: true, sounds good to me then

ohitstom:
I think before merging this we should wait 1 to 2 Spotify updates to see how much if at all the snackbar implementation will change? does that sound good to you? @rxri

rxri:
That's fine. I'm still looking into webpack hooks

ohitstom:
> That's fine. I'm still looking into webpack hooks

awesome, i have given it a few looks but couldnt find one that gave all the required information that the regex currently does :/

jsHelper/spicetifyWrapper.js
```diff
@@ -439,6 +447,26 @@ window.Spicetify = {
...
+	// Combine snackbar and notification
+	(async function bindShowNotification() {
+		if (!(Spicetify.Snackbar || Spicetify.sendNotification)) {
+			setTimeout(bindShowNotification, 10);
+			return;
+		}
+
+		Spicetify.showNotification = (message, isError = false, msTimeout) => {
+			if (!Spicetify.Snackbar?.enqueueSnackbar) {
+				Spicetify.notification(message, isError, msTimeout);
```
rxri, line 459: I forgot to change it here
```suggestion
				Spicetify.sendNotification(message, isError, msTimeout);
```

rxri (changes requested):

rxri (approved):
For the time being regex is fine. In the future I will try to move it to webpack and since they deleted it, it's probably gone but I will have a look for it later. I don't mind this pr being merged rn

ohitstom:
> For the time being regex is fine. In the future I will try to move it to webpack

so we are just gonna leave the old notification method in even if it's broken? for support reasons I suppose?

theRealPadster (approved):

rxri:
> > For the time being regex is fine. In the future I will try to move it to webpack
> 
> so we are just gonna leave the old notification method in even if it's broken? for support reasons I suppose?

Since spicetify supports version from 1.2.0 to "latest", we're going to keep this regex for older versions

afonsojramos (commented):
Nice!!

PR #2628, opened 2023-11-03
fix(wrapper): confirmdialog component

ohitstom:
New versions of Spotify do not select the correct function, whilst it would be preferable to use `functionModules.find(m => m.toString().includes("confirm-dialog-description"))` i do not have the patience to check every supported version of spotify to see if that classname is present. This approach should theoretically cover all older versions since kyries original pr showed onclose being passed as a prop.
```js
Spicetify.ReactDOM.render(
  Spicetify.ReactComponent.ConfirmDialog({
    titleText: "hi",
    descriptionText: "hi",
    cancelText: "hi",
    confirmText: "hi",
    confirmLabel: "hi",
  }),
  document.querySelector(".ReactModalPortal"),
);
```
image

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -384,7 +384,9 @@ window.Spicetify = {
 		ReactComponent: {
 			...Spicetify.ReactComponent,
 			TextComponent: modules.find(m => m?.h1 && m?.render),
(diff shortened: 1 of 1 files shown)


ohitstom:
this version handles states too unlike our current one

ohitstom:
@rxri sorry for the mention, any chance we can make a release once this fix is merged wanted to implement it into Comfy which i cant do until this is fixed?

afonsojramos (approved):
I suppose this fix should be enough to maintain compatibility, thanks!

rxri (approved):

PR #2647, opened 2023-11-09
feat(wrapper): append context menu item icons on the left

### LINKED ISSUE #2646: Update ContextMenu.Item to match Spotify

As of `Spotify for Windows (64 bit) 1.2.24.756.g7a7fc7f0` there are now official implementations of icons, i believe that context menu items should match this but we should leave `Spicetify.Menu.Item` with icons on the right.
image

However for `ContextMenu.Item` we should prepend the icons to match.

### ➕ Additional Information

_No response_

SunsetTechuila:
resolves #2646 
tested with spotify 1.2.24 and 1.2.0


SunsetTechuila:
tested with spotify 1.2.24 and 1.2.0

SunsetTechuila:
image
I didn't keep the profile menu icons on the right because icons which are passed to the Spicetify.(Context)Menu visually describe item text. The arrow icon of the "Account" item is on the right because it visually describes not the item text, but the action that will happen when you click it.
So if you want to follow Spotify's style - just don't use icons for the profile menu items

PR #2654, opened 2023-11-14
feat(wrapper): add unmapped svgicons to "misc" key

### LINKED ISSUE #2648: Many new icons to add to SVGIcons

As of `Spotify for Windows (64 bit) 1.2.24.756.g7a7fc7f0` there are many new icons thanks to the right click context menu, these need to be added or updated aswell as extensions such as trashbin.js having its trashbin icon updated to match spotify.

### ➕ Additional Information

image

ohitstom:
aims to resolve #2648
image

instead of misc?:
- unknown
- unmapped
- internal

I also want to start mapping 24px icons too but it may break compatibility to do:
```
icons {
    16: {},
    24: {}
}
```


ohitstom:
could use some help with this since its alot of manual labour if anyone wanted to pitch in

PR #2667, opened 2023-11-18
feat(preprocess): nuke ui logging

SunsetTechuila:
the `methodToDisable\{.+?\}nextMethodName` part is not very reliable and can potentially break things in the future, so improve it if know how. Or close the pull request if you don't know and are afraid of breakages, but I spent several hours on it


SunsetTechuila:
tested with these spotify versions
image

SunsetTechuila:
the reason I even try to replace all method contents is that `return' doesn't somehow stop execution
image

Delusoire:
It seems like all these methods use `eventSender.send` wouldn't it be better to just replace this one function with a noop?

SunsetTechuila:
> It seems like all these methods use eventSender.send wouldn't it be better to just replace this one function with a noop?

i'm already replacing it, but not all logging methods use eventSender.send, only the ones in `vendor~xpui.js`

> Or even make validateEventData return something non null.

makes the console filled with endless errors and doesn't prevent saving logging data into localStorage

rxri (approved):

afonsojramos (approved):

afonsojramos:
Nice! Thanks for the PR!

PR #2692, opened 2023-11-29
refactor(wrapper): use apis for `toggleHeart`, `getHeart`, `toggleMute`

SunsetTechuila:
tested with spotify 1.2.25 and 1.2.0
the `setHeart` method is new, so the docs need to be update

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -69,12 +69,15 @@ window.Spicetify = {
 		},
 		getMute: () => Spicetify.Player.getVolume() === 0,
 		toggleMute: () => {
-			document.querySelector(".volume-bar__icon-button").click();
+			Spicetify.Player.setMute(!Spicetify.Player.getMute());
 		},
 		setMute: b => {
-			const isMuted = Spicetify.Player.getMute();
-			if ((b && !isMuted) || (!b && isMuted)) {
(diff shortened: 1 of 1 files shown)


jsHelper/spicetifyWrapper.js
```diff
@@ -111,8 +114,12 @@ window.Spicetify = {
 		skipForward: (amount = 15e3) => {
 			Spicetify.Player.origin.seekForward(amount);
 		},
+		setHeart: b => {
+			const uri = [Spicetify.Player.origin._state.item?.uri];
+			b ? Spicetify.Player.origin._collection.add({ uri }) : Spicetify.Player.origin._collection.remove({ uri });
```
rxri, line 119: if you don't assign to the variable, don't use conditional ternary operator please
```suggestion
			if (b) Spicetify.Player.origin._collection.add({ uri })
			else Spicetify.Player.origin._collection.remove({ uri });
```

rxri (changes requested):

jsHelper/spicetifyWrapper.js
SunsetTechuila: you're just jealous of it coolness

rxri: xd

rxri (approved):

afonsojramos (approved):

PR #2701, opened 2023-12-04
feat: enhance snackbar exposition

SunsetTechuila:
removed private methods and props from the `Spicetify.Snackbar` object
added `SnackbarProvider` and `useSnackbar`


SunsetTechuila:
not sure where `useSnackbar` should be: in ReactHooks or in Snackbar

afonsojramos (approved):

PR #2717, opened 2023-12-10
fix(wrapper): don't assign `Player.data` if `item` is null

SunsetTechuila:


SunsetTechuila:
maybe increase the interval then?

SunsetTechuila:
image
it is not an issue with formatting

rxri (approved):

afonsojramos (approved):

PR #2719, opened 2023-12-11
feat(css-map): update classes

SunsetTechuila:

PR #2723, opened 2023-12-13
feat(workflows): use `dry` option for prettier action

SunsetTechuila:
https://github.com/spicetify/spicetify-cli/pull/2719#issuecomment-1851604429

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -16,6 +16,6 @@ jobs:
       - name: Prettify code
         uses: creyD/prettier_action@v4.3
         with:
-          prettier_options: --write .
-          commit_message: "style: prettify code"
-          github_token: ${{ secrets.SPICETIFY_GITHUB_TOKEN }}
+          prettier_options: --check .
+          only_changed: true
+          dry: true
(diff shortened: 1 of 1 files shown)


SunsetTechuila:
tested on my own fork

rxri (approved):

SunsetTechuila:
no, it still doesn't work https://github.com/SunsetTechuila/spicetify-cli/pull/3/

afonsojramos (approved):
The issue with this is that you need to ask people to run prettier on their own, which can be annoying in some cases. But at the same time I get the issue

SunsetTechuila:
> The issue with this is that you need to ask people to run prettier on their own

I actually found a way to fix committing, but I think auto-committing would be annoying for contributors (it annoyed me even when I was testing this change lol).
And you don't need to ask contributors for prettifying their code, since they get the notification about the workflow failure.
So I decided to leave this as a check, as it never worked as intended and everything was fine. But if you insist, I'll add auto-commit.

afonsojramos:
I don't insist really! But I remember that in the past people didn't prettify their code. But I think everyone is more aware now, so it should be fine!

SunsetTechuila:
Also, I suggest running the latest prettier with it's defaults on the whole repo again

rxri (approved):

PR #2725, opened 2023-12-15
refactor(cmd): rewrite `InitPaths` func

SunsetTechuila:
more readable code + no more "i switched to the classic version, but spicetify still says that i have the ms store one"

### DIFF
--- a/src/cmd/cmd.go
+++ b/src/cmd/cmd.go
@@ -61,68 +61,42 @@ func InitPaths() {
 	spotifyPath = utils.ReplaceEnvVarsInString(spotifyPath)
 	prefsPath = utils.ReplaceEnvVarsInString(prefsPath)
 
-	if len(spotifyPath) == 0 {
-		spotifyPath = utils.FindAppPath()
+	if _, err := os.Stat(spotifyPath); err != nil {
+		actualSpotifyPath := utils.FindAppPath()
 
-		if len(spotifyPath) == 0 {
+		if len(actualSpotifyPath) == 0 {
+			if len(spotifyPath) != 0 {
+				utils.PrintError(spotifyPath + ` does not exist or is not a valid path.`)
+			}
 			utils.PrintError(`Cannot detect Spotify location. Please manually set "spotify_path" in config-xpui.ini`)
 			os.Exit(1)
 		}
 
+		spotifyPath = actualSpotifyPath
 		settingSection.Key("spotify_path").SetValue(spotifyPath)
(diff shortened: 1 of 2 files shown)


src/cmd/cmd.go
```diff
@@ -61,68 +61,42 @@ func InitPaths() {
...
-	if len(spotifyPath) == 0 {
-		spotifyPath = utils.FindAppPath()
+	if _, err := os.Stat(spotifyPath); err != nil {
+		actualSpotifyPath := utils.FindAppPath()
 
-		if len(spotifyPath) == 0 {
+		if len(actualSpotifyPath) == 0 {
+			if len(spotifyPath) != 0 {
+				utils.PrintError(spotifyPath + ` does not exist or is not a valid path.`)
+			}
```
rxri, line 70:
```suggestion
				utils.PrintError(spotifyPath + ` does not exist or is not a valid path.`)
				os.Exit(1)
			}
```

rxri (changes requested):

src/cmd/cmd.go
SunsetTechuila: image
exit is after the next message

rxri: That's actually confusing, because detection is more as of automatic thing. I would separate these messages, the same goes to line [line 85](https://github.com/spicetify/spicetify-cli/pull/2725/files#diff-b6ea6e8b09ecc1b298f3af268cd468e956f7c16d2bdf19df8cdba790d405eaabR85)

SunsetTechuila: > because detection is more as of automatic thing
but it actually tries to detect

rxri: sure, but you then check the path inside config, and if it fails then this should fail without saying anything about detection

SunsetTechuila: again, the process:
1. it checks the path from the config
2. **if the check is failed, it tries to detect the path**
3. if it fails to detect the path, it shows the error "cannot detect the path"

are you **really** sure that it shouldn't show this error just because the path is set?

rxri: I understand the process of it, however I REALLY believe that it shouldn't show even if the path inside the config is invalid. If it can't detect it and path is invalid, then what's the point of showing this message. If you want to show this information, do it in different order but not at the same time

rxri (approved):

afonsojramos (approved):

PR #2744, opened 2023-12-31
fix(wrapper/ReactHook): make `DragHandler` findable again

rxri:


SunsetTechuila (approved):

afonsojramos (approved):

jsHelper/spicetifyWrapper.js
```diff
@@ -479,7 +479,7 @@ window.Spicetify = {
 			...Object.fromEntries(menus)
 		},
 		ReactHook: {
-			DragHandler: functionModules.find(m => m.toString().includes("data-dragging-uri")),
+			DragHandler: functionModules.find(m => m.toString().includes("setDropOriginUri") && m.toString().includes("data-dragging")),
```
afonsojramos, line 482: How long has this not worked? Should we add an `||` here?

SunsetTechuila: > How long has this not worked

it doesn't work only with the latest spotify

> Should we add an || here?

no, this works even with spotify 1.2.0

PR #2745, opened 2024-01-01
fix(wrapper/cards): use `functionModules` instead of `modules`

rxri:


theRealPadster (approved):
👍

CharlieS1103 (approved):
wow definitely checked this great work team 👍

Issue #2757, opened 2024-01-09
expose ability to assign context menus without react


ohitstom:
### 📝 Provide a description of the new feature

This is surely something spotify have a function for internally, the reason this would be beneficial is that i want to add a right click context menu item to a `Spicetify.Topbar.Button` which means i cant use a react component to render it since its using the spice api!

maybe theres something missing that i can do on my end?

### ➕ Additional Information
only current way im aware of
```js
return Spicetify.React.createElement(
	Spicetify.ReactComponent.ContextMenu,
	{
		offset: [0, 12],
		trigger: "click",
		placement: "top",
		menu: Spicetify.React.createElement(menuWrapper), // Spicetify.ReactComponent.Menu
	},
	Spicetify.React.createElement("div to assign to")
);
```

rxri:
Stop thinking that every component can be used without React
cc. https://github.com/spicetify/spicetify-cli/pull/2776#issuecomment-1904762028

ohitstom:
> Stop thinking that every component can be used without React cc. [#2776 (comment)](https://github.com/spicetify/spicetify-cli/pull/2776#issuecomment-1904762028)

whoa no need to be hostile, also this is quite literally the first issue ive opened with a request like this?

rxri:
Second one actually but whatever

PR #2776, opened 2024-01-22
feat: re-implement `ContextMenu` & `Menu`

Delusoire:
This is a fully backward-compatible rewrite of Context Menus, it doesn't rely on "injecting" HTML elements using Dom-manipulation or mutation observers.
Instead, it directly patches Spotify's react components for the context menu, at two different places, first to extract `[uris, uids, contextUri, ...` and expose them with a react context. And second, extend the `children` prop from the container.
This change makes the code significantly easier to understand, as it no longer relies on internal tippy features. It also fixes the long-standing issue of open Spotify sub-menus not closing after you hover over a Spicetify injected item.
image


Delusoire:
With the last commit, the Profile menu is also handled in the exact same way as the Context Menu.

src/preprocess/preprocess.go
```diff
@@ -369,6 +369,11 @@ func exposeAPIs_main(input string) string {
 		`\(\({[^}]*,\s*imageSrc`,
 		`Spicetify.Snackbar.enqueueImageSnackbar=${0}`)
 
+	// Menu hook
+	utils.Replace(&input, `children:([\w_$]+),onClose:[\w_$]+,getInitialFocusElement:[\w_$]+,onFocusVerticalItem:[\w_$]+.+?\}\)=>\{`, `${0}${1}=[Spicetify.ContextMenuV2.renderItems(),${1}].flat();`)
+
+	utils.Replace(&input, `\(0,([\w_$]+)\.jsx\)\([\w_$]+\.[\w_$]+,\{value:"contextmenu"[^\}]+\}\)\}\)`, `${1}.jsx((Spicetify.ContextMenuV2._context||(Spicetify.ContextMenuV2._context=i.createContext(null))).Provider,{value:{props:e?.props,trigger:l,target:p},children:${0}})`)
```
SunsetTechuila, line 375: **i**.createContext, **e**?.props, trigger:**l**, target:**p** ?

jsHelper/spicetifyWrapper.js
```diff
@@ -727,6 +729,8 @@ window.Spicetify = {
 		// isSameIdentity
 		Spicetify.URI.isSameIdentity = URIModules.find(m => typeof m === "function" && m.toString().match(/[\w$]+\.id===[\w$]+\.id/));
 	})();
+
+	document.dispatchEvent(new Event("webpack-loaded"));
```
SunsetTechuila, line 733: why is this not at the top of the function?

SunsetTechuila (commented):
I'm against having `ContextMenu` and `ContextMenuV2`

ohitstom:
may i ask if this rewrite will potentially solve this issue https://github.com/spicetify/spicetify-cli/issues/2757?
or will this rewrite only include adding of individual items to the context menu and not creating an entire context menu and assigning to a dom element?

SunsetTechuila:
> ContextMenuV2 is temporary and should replace ContextMenu & Menu in the future.

In other words, you said yourself that this is something for Spicetify v3. I agree, which is why I'm against it

>  this is a perfect use case for using events

This is perfect for finding out that everything has been successfully hooked except the thing you are using. Everything should be explicitly checked. Don't forget that Spicetify is a modding tool

src/preprocess/preprocess.go
Delusoire: This is now resolved, as I'm using regex to find those variable names.

SunsetTechuila:
> I'm against using webpack-loaded event

nvm, just rename it to `hooksDone` or something

Delusoire:
> > ContextMenuV2 is temporary and should replace ContextMenu & Menu in the future.
> 
> In other words, you said yourself that this is something for Spicetify v3. I agree, which is why I'm against it

This doesn't have to wait until V3, it is a drop-in replacement that works NOW. Why wait on a feature that doesn't break any existing API.

Delusoire:
> > I'm against using webpack-loaded event
> 
> nvm, just rename it to `hooksDone` or something

Are you referring to this commit: https://github.com/spicetify/spicetify-cli/pull/2776/commits/7ed19b975e4481ced58e3792276e135c514d2096 ?
I think that the current nomenclature is more appropriate. The loading is non-linear because we rely on `setTimeout` madness, and as such when `fireWebpackLoaded` other `Spicetify.*.*` props are still not set.

Loading and order of execution for the different helper scripts and extensions should be revised in V3 to reduce the async mess, and not have extension developers check for whatever `Spicetify.*.*` they're trying to access in a recursive `setTimeout`.

SunsetTechuila:
> nvm, just rename it to hooksDone or something

Now I'm against `onWebpackLoaded` hook. Everything should be explicitly checked. Don't forget that Spicetify is a modding tool so nothing is reliable

> Are you referring to this commit

No, I want the event to be renamed

> Why wait on a feature that doesn't break any existing API.

Because no one will use it until we force them to. And I don't see a reason to do that.

Actually, all the breaking changes can be done without releasing v3 if we don't touch the current code and just add the SpicetifyV3 object. /s

SunsetTechuila:
> > I'm against using webpack-loaded event
> 
> nvm, just rename it to `hooksDone` or something

No, as usual I got it right at once, but then mistakenly doubted myself. I'm against using the event too, since you don't check if `Spicetify.React` explicitly

Delusoire:
> > > I'm against using webpack-loaded event
> > 
> > 
> > nvm, just rename it to `hooksDone` or something
> 
> No, as usual I got it right at once, but then mistakenly doubted myself. I'm against using the event too, since you don't check if `Spicetify.React` explicitly

You're all over the place bro, I can't see where you stand.

So you'd rather have developers do:
image
than a simple:
`Spicetify.Hooks.onWebpackLoaded(() => ...)`

SunsetTechuila:
> than a simple: Spicetify.Hooks.onWebpackLoaded(() => ...)

as I said:

> Everything should be explicitly checked. Don't forget that Spicetify is a modding tool so nothing is reliable

If you want to implement something like this, at least:
1) Rename the hook and event
2) Dispatch the event only if `Spicetify.test()` is executed without any errors

rxri:
1. are all versions tested, 1.2.0 -> 1.2.29?
2. is the regex future-proof?
3. I don't like how you need to wrap stuff inside `Spicetify.Hooks.onWebpackLoaded(() => {})`
4. I agree with Grigory about: 
```
> Why wait on a feature that doesn't break any existing API.

Because no one will use it until we force them to. And I don't see a reason to do that.
```
no other questions atm

Delusoire:
> > than a simple: Spicetify.Hooks.onWebpackLoaded(() => ...)
> 
> as I said:
> 
> > Everything should be explicitly checked. Don't forget that Spicetify is a modding tool so nothing is reliable
> 
> If you want to implement something like this, at least:
> 
> 1. Rename the hook and event

There is no event.
The hook's name already reflects what it is / when it fires.

> 2. Dispatch the event only if `Spicetify.test()` is executed without any errors

And take a course on formal methods too? \s

`Spicetify.test()` SHOULD always run without errors, it is the maintainers responsibility to ensure that Spicetify's functionality is working, not the extension developer. The developer shouldn't care at all about whether this specific prop is loaded or not. The API SHOULD provide "signals" that indicate that all Network methods are loaded or all Webpack modules etc.

Spicetify.test() is only meant for testing environment / as a debugging tool. Not to be run for everyone on startup.

But if you insist on doing it your way, then you have to rethink how props are loaded. Probably something like: prop = assertNonNullAndReturn(<something>) and fire a groupA loaded event/callback at the end of the block, then try catch on the outer scope.

Delusoire:
> 4. I agree with Grigory about:
> 
> ```
> > Why wait on a feature that doesn't break any existing API.
> 
> Because no one will use it until we force them to. And I don't see a reason to do that.
> ```
> 
> no other questions atm




> > > Why wait on a feature that doesn't break any existing API.
> > 
> > 
> > Because no one will use it until we force them to. And I don't see a reason to do that.
> 
> Not if they would be already using it; ContextMenu & Menu are like adapters, you can use them the old way and they translate to the newer implementation.
> 
> Why don't you want to wait on it? Because this implementation:
> 
> > also fixes the long-standing issue of open Spotify sub-menus not closing after you hover over a Spicetify injected item.


this ^
This is fixing a bug. And the bug SHOULD be fixed without a need for a new major release that's probably getting more delays than nasa's Artemis program.

Delusoire:
> 3. I don't like how you need to wrap stuff inside `Spicetify.Hooks.onWebpackLoaded(() => {})`

And you'd rather have extension developers do this?
image
This is what they have to do in order to use 90% of the Spicetify wrapper objects.
And all this situation could have been prevented if the wrapper was split into two parts, and the other jsHelpers moved after the second part in `index.html`.
This is a symptom of how & when the jsWrappers are loaded, something that should definitely be reconsidered for V3.

SunsetTechuila:
> maintainers responsibility to ensure that Spicetify's functionality is working
 
But keeping spicetify installation updated is not

> There is no event

Oh, right, you've removed the event and added a non-react hook. Don't understand why

> The hook's name already reflects what it is / when it fires.

No, it is not. If it was, `Spicetify.Hooks.fireWebpackLoaded()` would be right after the `window.webpackChunkopen` check

Delusoire:
> 2. is the regex future-proof?

Regex and future-proof in one sentence?
Who knows?
Can you think of a better regex to do the job? Then you can use that one instead.

It relies on the fact that `value:"contextmenu"` is the first entry of props
And on the parameter names: menu, trigger, and triggerRef

So yeah i'd say it's pretty future proof

Delusoire:
> > The hook's name already reflects what it is / when it fires.
> 
> No, it is not. If it was, `Spicetify.Hooks.fireWebpackLoaded()` would be right after the `window.webpackChunkopen` check

Okay, so what do you want to name it? `fireAllSpicetifyPropsLoaded` ?
NO! Because at that point there is still a plethora of props that are STILL NOT loaded.

SunsetTechuila:
> Okay, so what do you want to name it?

https://github.com/spicetify/spicetify-cli/pull/2776#issuecomment-1904304685

Delusoire:
> > There is no event
> 
> Oh, right, you've removed the event and added a non-react hook. Don't understand why

First, hooks are in no way tied to react. They're a general concept that can be applied anywhere, and they predate react or any other js framework.

The event was simply an implementation detail.

Delusoire:
> > I'm against using webpack-loaded event
> 
> nvm, just rename it to `hooksDone` or something

Lmfao
Can you reread the full message that you replied to?
Naming it `hooksDone` would be LYING to the extension developers!!!
Not all hooks ARE DONE after hotloadWebpackModules is done!!!!!

Delusoire:
> > maintainers responsibility to ensure that Spicetify's functionality is working
> 
> But keeping spicetify installation updated is not

Wtf, where's the link between the two? An outdated installation would be running outdated spicetify & jsHelpers & extensions & spotify.

If Spotify releases a breaking change then extensions & jsHelpers will stop working anyway (even worse, with that setTimeout loop the extensions will keep wasting resources while not doing anything), and the user will be EXPECTED to update their installation.

Delusoire:
If you don't want to accept this change then close the pr. I don't like wasting time responding to the same question 3 times, and not getting any constructive feedback.

"Name it into something, but nothing of your choosing"

SunsetTechuila:
> An outdated installation would be running outdated spicetify & jsHelpers & extensions & spotify.

Read the support-tickets channel and you will see that a lot of people have the latest Spotify with outdated Spicetify

> If Spotify releases a breaking change then extensions & jsHelpers will stop working anyway 

Ofc not. Often when Spotify updates, something not vital is broken. Like `Locale`

Delusoire:
An extension developer can still check if Spicetidy.*.* is defined inside the callback still!!
The callback system is a mere improvement of the setTimeout recursive hell.

SunsetTechuila:
> An extension developer can still check if Spicetidy.. is defined inside the callback still!!

And use `setTimeout` if it is not..?

Delusoire:
> > An extension developer can still check if Spicetidy.. is defined inside the callback still!!
> 
> And use `setTimeout` if it is not..?

The step that's supposed to assign the prop failed...
That step is only done once...

ergo

The prop will never be assigned...

Extension developer can do as he pleases, but the prop will NEVER be available, they can be an idiot and fallback to a useless setTimeout, or define a fallback prop or panic or whatever.

Delusoire:
Here is a low quality gif demonstrating the bug:
gif

SunsetTechuila:
> The prop will never be assigned...

so all hooks are done after `hotloadWebpackModules` is done?

Delusoire:
> > The prop will never be assigned...
> 
> so all hooks are done after `hotloadWebpackModules` is done?

Lmfao, wtf man. Can't you take a look at spicetifyWrapper.js?

onLoadedWebpackModules's callbacks are called when all the Spicetify props set inside hotloadWebpackModules are set.
Not when ALL SPICETIFY PROPS ARE SET. How many times do I need to say that for you to understand...

Delusoire:
For example, Spicetify.Keyboard waits for Spicetify.Mousetrap which is exposed through webpackModules.
Reading Spicetify.Keyboard from inside a callback passed to onWebpackLoaded would always produce undefined.

SunsetTechuila:
I was speaking about the hooks all the time, not about _"all spicetify props"_ 

But ok, this dialog has clearly reached an impasse

To summarize:
1) I don't like having `ContextMenu` and `ContextMenuV2` at the same time. Already explained why: https://github.com/spicetify/spicetify-cli/pull/2776#issuecomment-1904338471
2) I want the hook to be renamed to `onWebpackHooksDone`

That's all **rn**

Delusoire:
> 1. I don't like having `ContextMenu` and `ContextMenuV2` at the same time. Already explained why: [feat: re-implement `ContextMenu` #2776 (comment)](https://github.com/spicetify/spicetify-cli/pull/2776#issuecomment-1904338471)

Solution that you proposed is simply to wait for V3 and release a breaking change.
My response was: This shouldn't have to wait because it's FIXING A BUG. And how stupid it is to hold off from fixing some bug because of a naming convention that can be settled in a matter of minutes. Also, V3 is nowhere near being production-ready, actually, V3 development has technically still not started (the V3 branch is like a tauri hello world) and no one seems to be motivated enough to pick it up.

> 2. I want the hook to be renamed to `onWebpackHooksDone`. Or would be even better to create a separate PR for that hook. But it is ok if you don't want to

Can't split the PR into two because the Menu changes make it depend on React, and the jsHelpers are run before Spicetify.React is loaded.
I'm not going to change the name of the dispatcher until everyone agrees on what to name it.

ohitstom:
> > > There is no event
> > 
> > 
> > Oh, right, you've removed the event and added a non-react hook. Don't understand why
> 
> First, hooks are in no way tied to react. They're a general concept that can be applied anywhere, and they predate react or any other js framework.
> 
> The event was simply an implementation detail.

would a non-react hook solve https://github.com/spicetify/spicetify-cli/issues/2757

Delusoire:
> would a non-react hook solve #2757 

Only way to use spotify's context menus is through the ReactComponent, so you have to use it. But if you don't wish to use react to generate the children dom tree then you have to use a useRef hook and create a div with react and ref passed to props. Then you can treat that ref as any other normal HTML element.

ohitstom:
> > would a non-react hook solve #2757
> 
> Only way to use spotify's context menus is through the ReactComponent, so you have to use it. But if you don't wish to use react to generate the children dom tree then you have to use a useRef hook and create a div with react and ref passed to props. Then you can treat that ref as any other normal HTML element.

ah that makes sense, thank you! ❤️

Delusoire:
Reopened as requested by @rxri.
The TLDR of the above discussion:
- The new implementation is more performant (as it doesn't rely on MutationObservers), and fixes 2 bugs, the first being:
    1. The incompatibility of Spotify's submenus with Spicetify-injected menu items.
    2. Spicetify's menu items were also injected into the add to playlist menu, triggered by clicking on the newly introduced plus sign.
- I sensed some confusion about the new ContextMenuV2:
    This pr exposes a new ContextMenuV2 object that serves as both an "infrastructure" and "preface" for the newer implementation, you can opt in to use it and get all the flexibility of modeling your UI with React directly (as if it was any other component written by Spotify as you get access to all the context providers etc...), or you can continue to use the older API, which internally translates every call to use ContextMenuV2 and get all the **bug fixes**.
- There was some backlash related to these new changes:
    1. "This could wait for V3":  We know neither when V3 is going to get released nor how the new jsHelpers are going to be reworked.
    2. `Spicetify.Hooks.onWebpackLoaded`: This merely reflects a "ready" state, or rather a "webpack related props (a.k.a: Props defined from webpack modules) should have been ready by now". It takes a callback and calls it back when initialization of webpack **related** props is finished. It can easily be turned into a Promise as follows: `await new Promise(resolve => Spicetify.Hooks.onWebpackLoaded(resolve))`, which is a much more succulent alternative to: 
    ```javascript
        function main() {
            if (!Spicetify.React || !Spicetify.ReactDOM || !Spicetify.Mousetrap || /* all other deps */) {
                setTimeout(main, /* wasted arbitrary waiting time */)
                return
            }
            /* .... */
        }
        main()```

jsHelper/spicetifyWrapper.js
```diff
@@ -1248,481 +1269,376 @@ Spicetify.SVGIcons = {
...
+			uris = props.uris;
+		} else if (props.uri) {
+			uris = [props.uri];
+		} else if (props.item?.uri) {
+			uris = [props.item.uri];
+		} else if (props.reference?.uri) {
+			uris = [props.reference.uri];
+		} else {
+			return;
+		}
```
afonsojramos, line 1290: Hate this as is, can we rewrite this whole function into these bits instead?
```suggestion
		const resolvedUri = props.uri || props.item?.uri || props.reference?.uri;
		const uris = props.uris || (resolvedUri ? [resolvedUri] : null);
		if (!uris) return;
```

jsHelper/spicetifyWrapper.js
```diff
@@ -1248,481 +1269,376 @@ Spicetify.SVGIcons = {
...
+		if (props.uids) {
+			uids = props.uids;
+		} else if (props.uid) {
+			uids = [props.uid];
+		} else if (props.item?.uid) {
+			uids = [props.item.uid];
 		}
```
afonsojramos, line 1297:
```suggestion
		const resolvedUid = props.uid || props.item?.uid;
		const uids = props.uids || (resolvedUid ? [resolvedUid] : undefined);
```

jsHelper/spicetifyWrapper.js
```diff
@@ -1248,481 +1269,376 @@ Spicetify.SVGIcons = {
...
-		if (realY + thisHeight > window.innerHeight) {
-			y -= realY + thisHeight - window.innerHeight;
-		}
-		this.style.transform = `translate(${x}px, ${y}px)`;
-	}
+		if (props.contextUri) {
+			contextUri = props?.contextUri;
+		} else if (props.context?.uri) {
+			contextUri = props.context.uri;
+		}
```
afonsojramos, line 1303:
```suggestion
		const contextUri = props?.contextUri || props.context.uri || undefined;
```

afonsojramos (changes requested):
Everything else looks alright, but I'll give it a re-review later

jsHelper/spicetifyWrapper.js
Delusoire: I'm not the original author of this function but I agree that it's not very readable as is. Though I have to ask, is there a specific reason for using the logical or over the nullish coalescing operator?

jsHelper/spicetifyWrapper.js
Delusoire: Wouldn't this change alter the behavior of this snippet? Shouldn't it be: `const contextUri = props.contextUri ?? props.context?.uri;` ? As:
1. props.context can be undefined
2. the original code doesn't check if props is undefined

jsHelper/spicetifyWrapper.js
afonsojramos: I think they have the same function here to be honest, both would result in the same functionality, although I feel like the logical OR is more commonly used, so it is more understandable.

jsHelper/spicetifyWrapper.js
afonsojramos: How so? 
Regarding 2 I didn't change that line: https://github.com/Delusoire/spicetify-cli/blob/master/jsHelper/spicetifyWrapper.js#L1276

Regarding 1, in the current code contextUri can also be undefined if you never define it if both conditionals fails.

Delusoire: I realize that my wording was ambiguous. What I tried to say is that the new snippet isn't faithful to the original. The optional chaining of `props?.contextUri` is unnecessary because `props` is never undefined. And second, if `props.context` is undefined then accessing `props.context.uri` will produce an error, this null check was previously done inside the else if's condition. Also, the ` || undefined` is not necessary either.

jsHelper/spicetifyWrapper.js
Delusoire: It's kind of odd not to pair `?.` with `??`, plus it's already been 4 years since the introduction of `??`. But if you don't want to use it that's not big of a deal

rxri: we should use `??` lol

afonsojramos (approved):

rxri (approved):

rxri:
merging to prod without testing xoxo /s

PR #2779, opened 2024-01-22
feat(vscode): add default formatters for other langs

rxri:


SunsetTechuila (changes requested):
I suggest setting biome as the formatter for js only, since it can't format yml and markdown

SunsetTechuila (approved):
> for js only
and for json too, right

afonsojramos (approved):

PR #2785, opened 2024-01-27
feat(install/unix): add spicetify to session PATH

rxri:


install.sh
```diff
@@ -114,7 +114,9 @@ check() {
                 echo >> "$shellrc"
             fi
             echo "${2:-$path}" >> "$shellrc"
-            log "Restart your shell to have spicetify in your PATH."
+            # Source the rc file to have spicetify binary available already in the session
+            # shellcheck source=/dev/null
```
SunsetTechuila, line 118:
```suggestion
            # shellcheck disable=SC1090
```

SunsetTechuila (dismissed):

install.sh
rxri: the SC1090 mentions that the `shellcheck source=/dev/null` disables it

install.sh
```diff
@@ -114,7 +114,9 @@ check() {
...
+            . "$shellrc"
```
SunsetTechuila, line 119: this does work
```suggestion
            export PATH="$PATH:$spicetify_install"
```

SunsetTechuila (changes requested):

install.sh
theRealPadster: I like this better ^

SunsetTechuila (approved):

afonsojramos (approved):

PR #2512, opened 2023-08-12
feat: toggle in-app update notification

ohitstom:
- removed disable_upgrade_check since it no longer works and is unlikely to be reintroduced
- set check_spicetify_upgrade default value to true
- added check_spicetify_upgrade to Spicetify.Config
- main feat: exit checkForUpdate() function if check_spicetify_upgrade is false (no update topbar button notification)

resolves #2507

PR #2789, opened 2024-01-30
feat: bring back option to disable spotify self-updates

SunsetTechuila:
re-implements the feature removed in #2512


rxri (commented):
image
this is my opinion about it to other reviewers btw

SunsetTechuila:
> because there's no easy way to restore upgrading functionality without restoring backup

Don't see any problem with this

> setting the config value to 0 will literally do nothing

Ofc, there is no setting that doesn't requires at least re-applying

PR #2804, opened 2024-02-02
feat(fullAppDisplay): enable all options by default

Ender-Wang:
- Whenever Spicetify gets an update, the `state` in `fullAppDisplay` seems lost, and I have to enable all those options **one by one every time**.
- For personal preference, I check all the options in `fullAppDisplay`, it's such a great extension, so why not enable it by default if someone like me clicks it and wanna the gorgeous experience of FULL app display mode?
- Now, the `state` is `true` by default. I am pretty sure the number of enabled options is greater than the number of disabled options if you wanna use `fullAppDisplay`.


rxri:
no. explanation is provided on the discord

PR #2829, opened 2024-02-14
fix(contextmenu/item): return to original `isEnabled` behavior

rxri:
That's how it was before...


theRealPadster (approved):

CharlieS1103 (approved):
Wow! Kudos to you, simply amazing work right here.

PR #2859, opened 2024-02-25
fix(lyrics-plus): disable genius for `1.2.31` and higher

rxri:
I don't have time, patience and knowledge to bypass cloudflare in a way, that genius would work all the time

### DIFF
--- a/CustomApps/lyrics-plus/Pages.js
+++ b/CustomApps/lyrics-plus/Pages.js
@@ -367,7 +367,7 @@ const SyncedExpandedLyricsPage = react.memo(({ lyrics, provider, copyright, isKa
 	const pageRef = useRef(null);
 
 	useTrackPosition(() => {
-		if (!Player.data.is_paused) {
+		if (!Spicetify.Player.data.is_paused) {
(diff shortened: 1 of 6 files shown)


theRealPadster (approved):

CharlieS1103 (approved):
LGTM

FiftyFour7250:
Can someone please explain why genius is working fine for me now on the old version of Spotify (1.1.62.583) and spicetify (2.5.0)? Yea, it's all very old, from the time when Spotify just introduced a new interface (2020 or 2021), and I periodically had to change something so that it would continue work, but I definitely didn't bypass Cloudfare lol. I don't think Cloudfire has anything to do with it, otherwise it wouldn't work on older versions either.

I can share this dinosaur if anyone needs it.

FiftyFour7250:
@rxri

rxri:
what? This has been disabled for 1.2.31 and higher because all requests from CosmosAsync go through our cors proxy which genius requires to work inside Spotify Client. My server has been flagged by Cloudflare as a bot, requiring me to pass captcha - which well, is not possible. Spotify had ability to do external requests to the network on versions before 1.2.31 that's why it's still available for these versions.

Issue #2877, opened 2024-03-03
lyrics-plus extreme cpu usage


nxtalex:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: newest
- Spicetify version: newest
```

### 📝 Description

pretty simple, lyrics plus uses 20% + cpu , that not good, it happened to a lot of other people as well so not my problem

rxri:
I think you're talking about Beautiful Lyrics, not lyrics-plus. Beautiful Lyrics is not official extension. It will use your CPU because Spotify has hardware acceleration disabled and animations, needs to be calculated somehow

nxtalex:
> I think you're talking about Beautiful Lyrics, not lyrics-plus. Beautiful Lyrics is not official extension. It will use your CPU because Spotify has hardware acceleration disabled, and animations needs to be calculated somehow

i am using full screen extension, and that extension usees lyrics plus extension, when i take it off it works fine. idk what to say 😕

rxri:
Problem with `lyrics-plus` is that it cannot use that much, by design. We don't even use `requestAnimationFrame` or `setTimeouts` to change UI. Generally, without any debug or performance logs I won't be able to say that it's lyrics-plus' fault since I cannot reproduce this issue

ameliorability:
Same issue, here are screenshots with CPU usage

With lyrics-plus DISABLED: 
![image](https://github.com/spicetify/spicetify-cli/assets/141889580/0edc1570-6ada-4cbf-b526-d057cb7acfe6)

With lyrics-plus ENABLED but not being used:
![image](https://github.com/spicetify/spicetify-cli/assets/141889580/726e7350-5271-4fbe-8326-2710e50cc741)

With lyrics-plus ENABLED and being used: 
![image](https://github.com/spicetify/spicetify-cli/assets/141889580/314ef314-5786-40d8-b2a4-4607841b0408)

But thank you rxri and other wonderful developers for this wonderful extension :)

rxri:
Like I said in my second message, I need performance logs or anything useful, since as mentioned in my previous comment, it's at around ~7% for me, so I cannot "debug" it. I'm sorry. I recommend trying when you don't have themes and extensions installed, and just lyrics-plus

SKDemon820:
[...]

PR #2882, opened 2024-03-07
fix(lyrics-plus): prevent re-rendering `TabBar` frequently

rxri:
since I was rewriting `links` variable, it always different every-time, causing the second `useEffect` to re-render this component very frequently. I changed it, and it should work correctly now.

I also optimized these functions because they took around ~600ms to execute anyway

fixes https://github.com/spicetify/spicetify-cli/issues/2877


nxtalex:
do i have to update the lyrics extension? howdo i do that?

PR #2888, opened 2024-03-10
fix: repair helpers to work on newer versions

rxri:
~~Fixing home config is not feasible because it requires multiple regexes to patch home~v2 thus we decided to remove it. It was broken for very long time anyway, and I'm 99% sure it doesn't work for versions we support~~


rxri:
* **#2888** <a href="https://app.graphite.dev/github/pr/spicetify/spicetify-cli/2888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @rxri and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/apply/apply.go
```diff
@@ -361,22 +348,6 @@ func insertCustomApp(jsPath string, flags Flag) {
...
 
-func insertHomeConfig(jsPath string, flags Flag) {
-	if !flags.HomeConfig {
-		return
-	}
-
-	utils.ModifyFile(jsPath, func(content string) string {
-		utils.ReplaceOnce(
-			&content,
-			`([\w$_\.]+\.sections\.items)(\.map)`,
```
afonsojramos, line 372: Is this the only regex for home config?

rxri: yes

afonsojramos: So, what do you mean when you say "it requires multiple regexes to patch home~v2"?

rxri: it would require multiple regexes to work on the versions we support

afonsojramos: I'd be more comfortable in marking this as help needed to see if anyone can help than simply removing the feature I think

rxri: well, okay... just saying that no one uses it i think because only one person in last few months asked about it 🤷‍♀️

theRealPadster (approved):

afonsojramos (approved):

PR #2895, opened 2024-03-11
ci: pin versions of actions for security reasons

rxri:
I believe that is a good security measure because tags can be replaced

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,19 +6,15 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-
     steps:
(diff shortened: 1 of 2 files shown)


theRealPadster (approved):
Sure? Makes it less readable, so I prefer the other way, but doesn't really matter too much.

afonsojramos (changes requested):
Yes, it is more secure, but let's not go this route, because then, updating these becomes a headache. We don't have that many security concerns, specially regarding github actions. So I'd prefer if we stayed as it is.

rxri:
Ok

PR #2896, opened 2024-03-11
ci: remove 'update the homebrew tap' step

rxri:
Spicetify has its own formula in `homebrew-core` repo, https://github.com/Homebrew/homebrew-core/blob/master/Formula/s/spicetify-cli.rb, so we don't really need to maintain our own tap for this. Anyone that uses our tap should switch to core tap

thanks to https://github.com/khanhas/homebrew-tap/pull/14, otherwise I wouldn't know that's the case tbh. 
We **should also archive** https://github.com/spicetify/homebrew-tap after merging this pull request

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,19 +98,6 @@ jobs:
         with:
           url: https://vps.itsmeow.dev/spicetify-update
           method: GET
-      - name: Update Homebrew tap
-        uses: mislav/bump-homebrew-formula-action@v3
-        if: ${{ !contains(github.ref, '-') }}
-        with:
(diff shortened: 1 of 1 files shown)


theRealPadster:
Hmm, when was that added? Does the core one auto update when Spicetify releases?

rxri:
> Hmm, when was that added? Does the core one auto update when Spicetify releases?

4 months ago, and they have their ci bot for that auto updating seems

theRealPadster:
> 4 months ago, and they have their ci bot for auto updating it seems

Dopeee

theRealPadster (approved):

afonsojramos (approved):

PR #2897, opened 2024-03-11
fix(homeConfig): support all versions

ohitstom:
in reference to [this description](https://github.com/spicetify/spicetify-cli/pull/2888#issue-2177763270)

regex tested on:
image

### DIFF
--- a/jsHelper/homeConfig.js
+++ b/jsHelper/homeConfig.js
@@ -11,11 +11,15 @@ SpicetifyHomeConfig = {};
 	const statusDic = {};
 	let mounted = false;
 
-	SpicetifyHomeConfig.arrange = sections => {
+	SpicetifyHomeConfig.arrange = container => {
+		let sections = container.sections?.items;
+
 		mounted = true;
 		if (list) {
-			return list;
+			container.sections.items = list;
+			return container;
 		}
+
 		const stickList = (localStorage.getItem("spicetify-home-config:stick") || "").split(",");
 		const lowList = (localStorage.getItem("spicetify-home-config:low") || "").split(",");
 		const stickSections = [];
@@ -41,7 +45,9 @@ SpicetifyHomeConfig = {};
 		sections = sections.filter(Boolean);
 
 		list = [...stickSections, ...sections, ...lowSections];
-		return list;
+		container.sections.items = list;
+
+		return container;
 	};
 
 	const up = document.createElement("button");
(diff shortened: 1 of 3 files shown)


Delusoire:
This doesn't allow rearranging Meet DJ and Video Coursed because they're only added later.
image

ohitstom:
> This doesn't allow rearranging Meet DJ and Video Courses because they're only added later.
> image
> 

this is still proof of concept and homeconfig has never been able to do that.

this approach is no different to the existing one except it means the Regex can be simpler once I've finished it.

rxri:
good shit, i tested it but it seems like the javascript side is broken now. When I click `down` or `up`, the whole section goes to bottom or top, it doesn't move one by one basically

ohitstom:
> good shit, i tested it but it seems like the javascript side is broken now. When I click `down` or `up`, the whole section goes to bottom or top, it doesn't move one by one basically

unless im misunderstanding i believe thats intended, basically stick puts it right at the top, lower puts it right at the bottom, then  up or down will move within either stuck sections or lowered sections - i did however notice that if you make a change, the second you change a page and come back it resets until your reload which it wasnt doing with my graphql approach?

rxri:
nvm i'm just stupid and it actually stays in the same place for me when i change pages

ohitstom:
> nvm i'm just stupid and it actually stays in the same place for me when i change pages

🎉 will do some testing on why thats not the case for me, might be user error or my localstorage key is messed up

rxri:
> thoughts on the menu item issue?

known, we didn't bother troubleshooting it further because it worked before and it broke somewhere idk

ohitstom:
> > thoughts on the menu item issue?
> 
> known, we didn't bother troubleshooting it further because it worked before and it broke somewhere idk

unfortunate, i had a quick look but the new system looks totally different to me haha
i did however fix the logic as its initial state was bugged (in homev2)

rxri (changes requested):
make `z-index` on `#spicetify-home-config` very high so it's always on top
run `go fmt .`
run biome plox

ohitstom:
> make `z-index` on `#spicetify-home-config` very high so it's always on top run `go fmt .` run biome plox

will do

rxri:
> > make `z-index` on `#spicetify-home-config` very high so it's always on top run `go fmt .` run biome plox
> 
> will do
> 
> edit: what is plox?

https://www.dictionary.com/e/slang/plox/
menu item checkmark will be fixed by https://github.com/spicetify/spicetify-cli/pull/2899

rxri (approved):

ohitstom:
> menu item checkmark will be fixed by https://github.com/spicetify/spicetify-cli/pull/2899

feel free to say no but may i request that icons still stay on the right like the account one>
image
then if item is enabled it replaces the given icon with a check? i feel like that matches Spotify a little bit more !

rxri:
The account one is like this because that's common to have icon like this on the right to external link, not on the left

ohitstom:
> The account one is like this because that's common or let's say "standard" to have icon like this on the right to external link, not on the left

sounds good just my OCD haha :)

afonsojramos (approved):

afonsojramos:
Let's gooooo

ohitstom:
> Can anyone explain what this home config feature actually does?

allows you to re-order cards in the home page

PR #2921, opened 2024-03-25
docs: create code_of_conduct.md

laignelet16:
I believe adding a code of conduct to this repository would allow the community to maintain a positive and friendly environment and establish expectations for behavior for the project’s participants. If there is a specific contact method that I could add to the code of conduct that would be great! Let me know if any changes are needed.


rxri:
We are here for a good time, and joke around about stuff that are listed in this CoC. We're just fun, experimental project around spotify client, so it doesn't need to be that serious. If someone will constantly be a bad person we will punish them accordingly

PR #2928, opened 2024-03-30
ci(release): add support for windows `arm64`

rxri:
i don't kinda understand why we don't support windows arm64. Spotify supports it, so why won't we?


theRealPadster (approved):

afonsojramos:
Let's see if we get any weird issue reports 🧐

rxri:
Well it built on my machine, I can't run it ofc but it should work

PR #2932, opened 2024-03-31
fix(install/linux): properly parse `--root` flag and tag option

rxri:


afonsojramos (changes requested):
What is this flag for? We have no documentation for it anywhere.

rxri:
> What is this flag for? We have no documentation for it anywhere.

@afonsojramos how is it not... 
image

afonsojramos:
Well, fair. I just meant that we don't show it in any `--help` command, and that we don't have it in the docs. But I suppose that for this case that is sort of the goal, as it is an exception to the rule.

rxri:
Well yeah, it's intended tbh.

install.sh
```diff
@@ -5,13 +5,23 @@
 
 set -e
 
+for arg in "$@"; do
+    shift
+    case "$arg" in
+        "--root") set -- "$@" "-r" ;;
+        *) set -- "$@" "$arg"
+    esac
+done
```
afonsojramos, line 14: So all we're doing is to change the --root flag into -r?

install.sh
```diff
@@ -5,13 +5,23 @@
...
+
 while getopts ":r" arg; do
     case "${arg}" in
         "r") override_root=1 ;;
```
afonsojramos, line 18: Couldn't we just change this to `root`?

install.sh
rxri: yeah

install.sh
rxri: well, most of scripts/cli provide short version an long version so that's why it has both 🤷‍♀️

afonsojramos: Yeah, but that doesn't always happens as it is a rule that doesn't apply to all flags. In this case, we are talking about a flag that shouldn't be used outside of exceptions, so I would say that we shouldn't have a short version for this one, and just simplify and only accept `--root`

rxri: alright

afonsojramos (approved):

theRealPadster (approved):

PR #2956, opened 2024-04-08
feat(windows/install): add clink completions

Bergbok:
Add help text for when the user inputs: ?

image

Add clink completions for Cmder users

lil showcase:
GIF 08-04-2024 18-50-02

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -167,7 +167,11 @@ if (-not (Test-Admin)) {
   Write-Unsuccess
   Write-Warning -Message "The script was ran as Administrator which isn't recommended"
   $Host.UI.RawUI.Flushinputbuffer()
-  $choice = $Host.UI.PromptForChoice('', 'Do you want to abort the installation process to avoid any issues?', ('&Yes', '&No'), 0)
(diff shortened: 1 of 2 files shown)


rxri:
Well, this just sounds like a bloat that would be added to repo. Sorry, but I don't think it's a good idea to merge it within the main cli repo. I recommend just doing separate winget package for it for example.
About the choice descriptions, it's intended to not be easy for end-user because you ideally shouldn't install spicetify in Administrator window. If someone is able to understand questions like this, they will get through this.
thanks for the pr though

Bergbok:
I don't think it's bloat, simply a nice feature that very few users will even see since most won't have Cmder installed.
If it's not applicable to the user, it won't prompt.
`if ($null -ne $env:CMDER_ROOT)`
Those that use Cmder will enjoy love it though.

Regarding the help text - it's not only for when you run run it as admin, it's for all PromptForChoice calls.
Having it there is undeniably better than not.

Currently:
image

rxri:
Make the separate pull request just for help text, about cmder - look on #dev-off-topic on discord server because I talked about it.

PR #2982, opened 2024-04-24
fix(main-homeHeader): add new prop

sanoojes:
I think its changed


rxri (changes requested):
Do not delete the old one. Add it as a new mapping

sanoojes:
> You do not delete the old one. Add it as a new mapping

image

what is this one called it behaves same as "main-home-homeHeader" 

I'm a newbie, still learning. Sorry if I cause any inconvenience.

PR #2984, opened 2024-04-25
fix(lyrics-plus): force assign `currentLyrics` for cache

Delusoire:
…ranslateLyrics when hitting CACHE

Only fixes it when hitting CACHE, translateLyrics can *in theory* still be executed with outdated this.state.currentLyrics if tryServices resolves before the next component's rerender.


rxri:
Oh man

rxri (approved):

PR #3011, opened 2024-05-07
fix(wrapper/platform): assign product state to old prop

rxri:
fixes productstate on `1.2.37`

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -936,14 +936,14 @@ Spicetify.Events = (() => {
 
 	(function waitProductStateAPI() {
 		if (!Spicetify.Platform?.UserAPI) {
-			setTimeout(waitUserAPI, 1000);
+			setTimeout(waitUserAPI, 100);
 			return;
 		}
 
 		const productState = Spicetify.Platform.UserAPI?._product_state || Spicetify.Platform.UserAPI?._product_state_service;
 		if (productState) return;
 		if (!Spicetify.Platform?.ProductStateAPI) {
-			setTimeout(waitProductStateAPI, 1000);
+			setTimeout(waitProductStateAPI, 100);
 			return;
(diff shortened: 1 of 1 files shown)


theRealPadster (approved):

jsHelper/spicetifyWrapper.js
```diff
@@ -921,6 +934,23 @@ Spicetify.Events = (() => {
 		playerState.cache = playerState.current;
 	});
 
+	(function waitProductStateAPI() {
+		if (!Spicetify.Platform?.UserAPI) {
+			setTimeout(waitUserAPI, 100);
```
Delusoire, line 939: waitUserAPI is called regardless of this, so is it really needed here?

jsHelper/spicetifyWrapper.js
```diff
@@ -921,6 +934,23 @@ Spicetify.Events = (() => {
...
+			return;
+		}
+
+		const productState = Spicetify.Platform.UserAPI?._product_state || Spicetify.Platform.UserAPI?._product_state_service;
```
Delusoire, line 943: The `if (!Spicetify.Platform?.UserAPI) { /* ... */; return }` check prevents `Spicetify.Platform.UserAPI` from ever being null or undefined, so the optional chaining is not needed.

rxri: Not valid. The prop can be undefined so it's better to stay like this

Delusoire: `a?.b` is equivalent to `a === null || a === undefined ? undefined : a.b`, b is irrelevant from the operator's perspective. The fact that `_product_state` and `_product_state_service` can be undefined changes nothing.

rxri: right... mb

Delusoire (approved):

PR #3018, opened 2024-05-10
fix(lyrics-plus/netease): check correctly for songs

dompling:
lyrics time matching logic（For example:给我一首歌的时间）lyrics content mismatch

### DIFF
--- a/CustomApps/lyrics-plus/ProviderNetease.js
+++ b/CustomApps/lyrics-plus/ProviderNetease.js
@@ -19,7 +19,16 @@ const ProviderNetease = (() => {
 		// normalized expected album name
 		const neAlbumName = Utils.normalize(info.album);
 		const expectedAlbumName = Utils.containsHanCharacter(neAlbumName) ? await Utils.toSimplifiedChinese(neAlbumName) : neAlbumName;
-		const itemId = items.findIndex(val => Utils.normalize(val.album.name) === expectedAlbumName || Math.abs(info.duration - val.duration) < 1000);
+		let itemId = items.findIndex(val => Utils.normalize(val.album.name) === expectedAlbumName);
+
+        if (itemId === -1) {
+            itemId = items.findIndex(val => Math.abs(info.duration - val.duration) < 3000);
+        }
+
+        if (itemId === -1) {
+            itemId = items.findIndex(val => val.name === cleanTitle);
+        }
+
 		if (itemId === -1) throw "Cannot find track";
 
 		return await Spicetify.CosmosAsync.get(lyricURL + items[itemId].id, null, requestHeader);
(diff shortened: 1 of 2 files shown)


rxri (changes requested):
Change your code to align with biome errors. Also run biome linter on it too.

CustomApps/lyrics-plus/ProviderNetease.js
```diff
@@ -19,7 +19,16 @@ const ProviderNetease = (() => {
...
+		let itemId = items.findIndex(val => Utils.normalize(val.album.name) === expectedAlbumName);
+
+        if (itemId === -1) {
+            itemId = items.findIndex(val => Math.abs(info.duration - val.duration) < 3000);
+        }
+
+        if (itemId === -1) {
+            itemId = items.findIndex(val => val.name === cleanTitle);
+        }
+
```
rxri, line 31:
```suggestion
		const itemId = items.findIndex(val => Utils.normalize(val.album.name) === expectedAlbumName) ?? (items.findIndex(val => Math.abs(info.duration - val.duration) < 3000) || items.findIndex(val => val.name === cleanTitle));
```

rxri (changes requested):

CustomApps/lyrics-plus/ProviderNetease.js
rxri, line 31:
```suggestion
		const itemId = items.findIndex(val => 
			Utils.normalize(val.album.name) === expectedAlbumName ||
    		Math.abs(info.duration - val.duration) < 3000 ||
    		val.name === cleanTitle
		);

```

dompling: Thank you for your reply, but there may be an album mismatch, and then there may be a market match for that data, but it's not actually the original lyrics, so it's more accurate to iterate over it

rxri: What you did, with these two ifs, can be shortened into the change I proposed. There's no difference - just better looking code

dompling: Sure, the translation of the provided code snippet would be:
```javascript
(Utils.normalize(val.album.name) === expectedAlbumName ||
    Math.abs(info.duration - val.duration) < 3000 ) && val.name === cleanTitle

```

rxri: Then just do how you got it working, however do it inside the `items.findIndex` and not separate if statements

dompling: Thanks for your help

rxri (approved):

PR #3019, opened 2024-05-10
fix(keyboardShortcut): correctly call functions for volume

senisia:
changed the keybinding ctrl+up to increase the volume instead of decrease


afonsojramos (approved):
The fact that no one else noticed this shows how this is not that used 😅

ohitstom:
> The fact that no one else noticed this shows how this is not that used 😅

why do you think i removed it a couple prs ago and was reluctant to re-add it haha

regardless mb for missing this!

rxri (approved):

PR #3033, opened 2024-05-20
feat: drop `spicetify-` from name & move to `main`

afonsojramos:


rxri (approved):
Seems fine but test is needed if everything builds

rxri:
But wrapper needs a change as well iirc

afonsojramos:
> why renaming it to cli if its going to be a tui 😨

we're not renaming, we're just removing the `spicetify-` from the name

Delusoire (approved):
Looks all good👍

rxri (approved):

PR #3036, opened 2024-05-20
chore: bump biome schema

afonsojramos:


.vscode/settings.json
```diff
@@ -6,7 +6,13 @@
...
 		"editor.defaultFormatter": "ms-vscode.powershell"
 	},
-	"[javascript][typescript][json]": {
+	"[javascript]": {
+		"editor.defaultFormatter": "biomejs.biome"
+	},
+	"[typescript]": {
+		"editor.defaultFormatter": "biomejs.biome"
+	},
+	"[json]": {
```
rxri, line 15: why tho? the grouping works tho

afonsojramos: It is, indeed, but I keep having issue with this. But you're right, I'll rever!

rxri (approved):

PR #3051, opened 2024-05-28
chore(install): change wording for admin/root

rxri:
To make users more aware perhaps?


afonsojramos (approved):

afonsojramos:
Looks good, yeah, don't see why not!

PR #3087, opened 2024-06-29
ci(build): update `winget-releaser` to `latest`

vedantmgoyal9:


rxri (approved):
Oh yea it would be nice because v2 with the `x` version of komac didn't pick up arm64 builds

PR #3103, opened 2024-07-13
feat(css-map): map global nav active button

sanoojes:
active class for browse button

Screenshot 2024-07-13 134107


rxri (approved):

rxri:
I have never mapped this button because it's not necessary but whatever

PR #3109, opened 2024-07-16
Fix(sidebarConfig): The sidebar config menu was incorrectly enabled on the global nav.

sanoojes:
This fixes the side bar config menu incorrectly enabled on global nav now the menu shows only in libraryx and legacy 

context menu after global nav: 
image

context menu before global nav: 
image

context menu at library x remains same:
image


rxri (commented):
I don't see any reason why would it happen. It would make people think that it's not enabled in their config. Sidebar config needs to support global navbar, and not to remove its item from the Context Menu entirely

PR #3112, opened 2024-07-21
feat: parallel workflows, cors proxy

Delusoire:


build/macos/build.sh
```diff
@@ -3,30 +3,44 @@ set -e
...
+# plutil -remove NSAppleMusicUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSCalendarsUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSCameraUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSContactsUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSHomeKitUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSMicrophoneUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSPhotoLibraryUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSRemindersUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSSiriUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
+# plutil -remove NSSystemAdministrationUsageDescription ./Volume/Spicetify.app/Contents/Info.plist
```
rxri, line 32:
```suggestion
```

cmd/daemon.go
```diff
@@ -97,7 +104,7 @@ func startDaemon() {
 		}
 		defer watcher.Close()
 
-		log.Println("Watching:", paths.GetSpotifyAppsPath(spotifyDataPath))
+		log.Println("Watcher: watching:", paths.GetSpotifyAppsPath(spotifyDataPath))
```
rxri, line 107: maybe `[Watcher]`? Or perhaps leave it and I will make log function for every print/errors etc if you don't mind

build/macos/build.sh
```diff
@@ -3,30 +3,44 @@ set -e
...
+
```
rxri, line 33:
```suggestion

```

cmd/daemon.go
Delusoire: Let's not make this PR bigger than it needs to be, we can discuss how to do logging separately.

PR #3124, opened 2024-08-02
feat(update): run `backup apply` for stock spotify

rxri:
The reason for this change is that people assume spicetify has an update when spotify is in stock. This will only work when spotify is backupable. This change was voted positively in spicetify server.
image

### DIFF
--- a/spicetify.go
+++ b/spicetify.go
@@ -232,14 +232,14 @@ func main() {
 	utils.PrintBold("spicetify v" + version)
 	if slices.Contains(commands, "upgrade") || slices.Contains(commands, "update") {
 		updateStatus := cmd.Update(version)
-		if updateStatus {
-			ex, err := os.Executable()
-			if err != nil {
-				ex = "spicetify"
-			}
+		spotifyPath := filepath.Join(utils.FindAppPath(), "Apps")
+		spotStat := spotifystatus.Get(spotifyPath)
+		ex, err := os.Executable()
+		if err != nil {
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):

afonsojramos:
That's fair, they're already running a command either way so that means something. The text could be improved I think 🤔

spicetify.go
```diff
@@ -251,6 +251,15 @@ func main() {
 			cmd = exec.Command(ex, strings.Join(commands[:], " "))
 			utils.CmdScanner(cmd)
 		}
+
+		if spotStat.IsBackupable() {
+			utils.PrintInfo("spicetify is already up-to-date! If you ran this command because spicetify disappeared after Spotify updated, we'll attempt to fix it for you right now.")
```
afonsojramos, line 256:
```suggestion
			utils.PrintInfo("There was an update to Spotify, but Spicetify is still up-to-date, so we'll try to reapply Spicetify for you!")
```
What do you think?

rxri: We don't know if there was an update, i think we can reword the message I made but not change the meaning of it, since your change changes it in some way

afonsojramos: You think? I'm still saying that Spicetify is no longer applied, but it is up to date 🤔

rxri: Sure, but I always use the word "if". You're making a point in this message even if it could be not true. Also, we're attempting to fix it because we have no way if it will work or not

PR #3127, opened 2024-08-06
fix(new-releases): use `getContents` for `getArtistsList`

machinemessiah:
Restore `getArtistList` functionality by using `LibraryAPI.getContents` to perform the same operation that was previously handled by `LibraryAPI.getArtists` which no longer exists.

### DIFF
--- a/CustomApps/new-releases/index.js
+++ b/CustomApps/new-releases/index.js
@@ -274,9 +274,8 @@ async function getArtistList() {
 		sortOrder: ["0"],
 		textFilter: "",
 		offset: 0,
-		limit: 0,
+		limit: (await Spicetify.Platform.LibraryAPI.getContents(config))?.totalLength ?? 0,
 	};
-	config.limit = (await Spicetify.Platform.LibraryAPI.getContents(config)).totalLength;
 	const artists = await Spicetify.Platform.LibraryAPI.getContents(config);
 	count(true);
 	return artists.items ?? [];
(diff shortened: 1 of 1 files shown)


machinemessiah (commented):
Could be cleaner, but it works.

CustomApps/new-releases/index.js
```diff
@@ -269,8 +269,15 @@ class Grid extends react.Component {
...
-	const base = await Spicetify.Platform.LibraryAPI.getArtists();
-	const artists = await Spicetify.Platform.LibraryAPI.getArtists({ limit: base.totalLength });
+	const config = {
+		filters: ["1"],
+		sortOrder: ["0"],
+		textFilter: "",
+		offset: 0,
+		limit: 0,
+	};
+	config.limit = (await Spicetify.Platform.LibraryAPI.getContents(config)).totalLength;
```
rxri, line 279:
```suggestion
		limit: (await Spicetify.Platform.LibraryAPI.getContents(config))?.totalLength ?? 0,
	};
```

rxri (changes requested):

CustomApps/new-releases/index.js
machinemessiah: Cleaner for sure!
Not sure why I didn't do that to begin with...

Delusoire:
and where are the comments mate?

rxri:
Maybe post your review first then ask where are the comments. I already told you so in discord dms.

PR #3131, opened 2024-08-08
fix(lyrics-plus): remove `encodeURIComponent` from `URLSearchParams` in genius provider

irlbunny:
Fixes Genius (at least on 1.2.13)


Delusoire (approved):
looks good, URLSearchParams already url-encodes the search parameters

rxri (approved):

PR #3143, opened 2024-08-25
fix(wrapper): wait until all modules loaded

kyrie25:
Suggested solution for waiting for lazy-loaded modules. If there are required changes, please commit them directly onto the branch

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -185,6 +185,7 @@ window.Spicetify = {
 					"ContextMenuV2",
 					"ReactJSX",
 					"_renderNavLinks",
+					"Events",
 				]),
 			},
 			{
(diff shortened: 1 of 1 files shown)


jsHelper/spicetifyWrapper.js
```diff
@@ -348,7 +349,7 @@ window.Spicetify = {
 })();
 
 (function addProxyCosmos() {
-	if (!Spicetify.Player.origin?._cosmos && !Spicetify.Platform?.Registry) {
+	if (!Spicetify.Player.origin || !Spicetify.Platform?.Registry) {
```
rxri, line 352: That was intentional tho

rxri (changes requested):

jsHelper/spicetifyWrapper.js
kyrie25: I get the idea was to wait for either `_cosmos` or `Registry`, but in this case `origin` may not be available before `Registry` which causes a TypeError when assigning `_cosmos` to `origin`
https://github.com/spicetify/cli/blob/eaa16bb2aed8f30bcd481aeac9af4ec66585a0eb/jsHelper/spicetifyWrapper.js#L440

rxri: When did you get this issue because I have never seen TypeError in this instance

rxri (approved):

PR #3161, opened 2024-09-08
fix(home-config): support up to 1.2.45

ohitstom:
tested on 1.2.45, 1.2.44, and 1.2.41

### DIFF
--- a/src/preprocess/preprocess.go
+++ b/src/preprocess/preprocess.go
@@ -556,6 +556,14 @@ func exposeAPIs_main(input string) string {
 			return fmt.Sprintf("Spicetify.Snackbar.enqueueImageSnackbar=%s", submatches[0])
 		})
 
+	// Home Config >= 1.2.45
+	utils.ReplaceOnce(
+		&input,
+		`(&&"HomeShortsSectionData".*\],)([a-zA-Z])(\}\)\()`,
+		func(submatches ...string) string {
+			return fmt.Sprintf("%sSpicetifyHomeConfig.arrange(%s)%s", submatches[1], submatches[2], submatches[3])
+		})
+
 	// Menu hook
 	utils.Replace(&input, `("Menu".+?children:)([\w$][\w$\d]*)`, func(submatches ...string) string {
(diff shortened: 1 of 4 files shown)


src/preprocess/preprocess.go
```diff
@@ -556,6 +556,14 @@ func exposeAPIs_main(input string) string {
 			return fmt.Sprintf("Spicetify.Snackbar.enqueueImageSnackbar=%s", submatches[0])
 		})
 
+	// Home Config >= 1.2.45
+	utils.ReplaceOnce(
```
rxri, line 560: put it in the `apply.go`, add another line in the `filesToModified`:
```diff
- filepath.Join(appsFolderPath, "xpui", "home-v2.js"):             insertHomeConfig,
+ filepath.Join(appsFolderPath, "xpui", "home-v2.js"):             insertHomeConfig,
+ filepath.Join(appsFolderPath, "xpui", "xpui.js"):                insertHomeConfig,
```
and then just put it in the if statement on the line 351

ohitstom: why would i put it in the if statement? can i just 
```js
func insertHomeConfig(jsPath string, flags Flag) {
	if !flags.HomeConfig {
		return
	}

	utils.ModifyFile(jsPath, func(content string) string {
		utils.ReplaceOnce(
			&content,
			`(createDesktopHomeFeatureActivationShelfEventFactory.*?)([\w\.]+)(\.map)`,
			func(submatches ...string) string {
				return fmt.Sprintf("%sSpicetifyHomeConfig.arrange(%s)%s", submatches[1], submatches[2], submatches[3])
			})

		// >= 1.2.45
		utils.ReplaceOnce(
			&content,
			`(&&"HomeShortsSectionData".*\],)([a-zA-Z])(\}\)\()`,
			func(submatches ...string) string {
				return fmt.Sprintf("%sSpicetifyHomeConfig.arrange(%s)%s", submatches[1], submatches[2], submatches[3])
			})
		return content
	})
}
```

rxri (approved):

Issue #3176, opened 2024-09-23
Enable enter on search


asgerkromand:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: 1.2.45.454.gc16ec9f6
- Spicetify version: v2.38.3
```

### 📝 Description

I use the keyboardshortcut extension, and I'm trying to use the 'F'-key aka. keyboard-driven navigation. The issue happens when I navigate to the search bar and then type in e.g. an artist's name. 

Here I would then like to be able to press 'F' and jump to a given result from the search, but this is not possible as I'm still in the search bar. I then try to press Esc to get out of the search bar, but this deletes the search-keywords and the search results disappear. If instead I try to press enter to get out of the search bar, while keeping the search-keywords - and thus the search result - this is not possible neither.

In sum, I reckon it is not possible to use 'F' to get to the results without using e.g. tab. Could it be possible to add the 'Enter' as a possible key to press to get out of the search bar while keeping your search results, so that you can use the 'F'-key to go into the results.

### 📸 Screenshots

<img width="985" alt="billede" src="https://github.com/user-attachments/assets/ff24570b-d31f-4d3c-9164-87c601099153">

UmairRaaz:
I want to work in this

AishwaryaAgr:
Please review PR with the bug fixed

rien333:
> Here I would then like to be able to press 'F' and jump to a given result from the search, but this is not possible as I'm still in the search bar. I then try to press Esc to get out of the search bar, but this deletes the search-keywords and the search results disappear.

I think this can be closed for now, because this is no longer how spotify (`v1.2.47.364.gf06e5cee`) behaves. This is probably because the search functionality has been revamped.

Specifically, the following steps should do what OP wants:

1. Start a search, either by clicking the search field, or pressing <kbd>/</kbd> (when https://github.com/spicetify/cli/issues/3200 is resolved)
2. Press <kbd>ESC</kbd> twice
3. Press <kbd>f</kbd> to start hinting (that's what it's properly called, not "Vim mode", as it says in the source code)

rxri:
When you will click enter while search bar is focused, it will stop being focused and you will be able to use vim keybinds.

PR #3192, opened 2024-10-01
fix: issue 3176 fixed to Enable Enter on Search in keyboard shortcut extension

AishwaryaAgr:
Issue:
#3176 Enable Enter on Search

Description:
This pull request introduces a new feature that allows the focus to change when the user presses the "Enter" key while an input field is active in the keyboardShortcut.js. This enhancement enables users to use the navigation by clicking "f" to control the app.

Changes:

Added an event listener for the "Enter" key press on input fields.
When "Enter" is pressed, the focus is removed from the input field.
Ensured the default action of the "Enter" key is prevented to avoid unwanted form submissions.

Testing:

Tested the functionality in multiple browsers to verify consistent behavior.
Verified that the focus correctly shifts when pressing "Enter" on the active input.


rxri:
That's not how it should be fixed though. You're listening for this event on whole document which shouldn't be a thing.

PR #3237, opened 2024-12-05
Added option to explicitly specify the height and width of the canvas instead of using aspect ratio. Added more font size options. Also added option to specify horizontal and vertical padding

sarthak9793:


rxri (changes requested):
I can't read anything of it because of the formatting. Run biome.

rxri:
Closing due to inactivity.

Issue #3122, opened 2024-08-02
Alt+Left Arrow shortcut doesn't work in the 'Keyboard Shortcut' extension.


sevennotfound:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: 1.2.42.290.g242057a2 
- Spicetify version: v2.36.15
```

### 📝 Description

Pressing Alt + Left Arrow key is supposed to change the current song to the previous song (according to the code), but that doesn't work.

rxri:
It's `ctrl + left arrow`.

sevennotfound:
> It's `ctrl + left arrow`.

sorry it was a typo. ctrl + left arrow also doesn't work. I checked again

rxri:
> > It's `ctrl + left arrow`.
> 
> sorry it was a typo. ctrl + left arrow also doesn't work. I checked again

Is your spotify window focused? because it needs to be. I also checked it, and works fine on my end

sevennotfound:
> > > It's `ctrl + left arrow`.
> > 
> > 
> > sorry it was a typo. ctrl + left arrow also doesn't work. I checked again
> 
> Is your spotify window focused? because it needs to be. I also checked it, and works fine on my end

Yes it is focused and it still doesn't work. I'm on archlinux with gnome wayland if it helps.

rxri:
> > > > It's `ctrl + left arrow`.
> > > 
> > > 
> > > sorry it was a typo. ctrl + left arrow also doesn't work. I checked again
> > 
> > 
> > Is your spotify window focused? because it needs to be. I also checked it, and works fine on my end
> 
> Yes it is focused and it still doesn't work. I'm on archlinux with gnome wayland if it helps.

Spotify doesn't fully support wayland. Use `xwayland` to launch spotify, and it should work, if not, then it's just because of no support from Spotify side

PR #3274, opened 2025-01-24
fix: Update keyboardShortcut.js

Sk16er:
# issue
 - The shortcut for switching to the previous track in keyboardShortcut.js incorrectly calls Spicetify.Player.prev, a non-existent function, instead of Spicetify.Player.back.

Solves #3122, which was closed without proper testing of the extension.


rxri:
Duplicate of #3274 3273

Issue #3201, opened 2024-10-12
`spicetify/Backup` should not reside in `~/.config` | Or: following the XDG spec


rien333:
### 📝 Provide a description of the new feature

`~/.config` is meant for configuration files. I think anything that customizes spotify in any way should be stored here, as is mostly the case today.

There's a couple of exceptions, however: by default, spicetify stores Spotify backups in `~/.config/spicetify/Backup`, and extracted files in `~/.config/spicetify/Extracted`. 

This is both conceptually wrong, and practically annoying. The current behavior is conceptually wrong because spotify backups/extractions are not configuration files. [Looking at the XDG spec](https://specifications.freedesktop.org/basedir-spec/latest/), `$XDG_STATE_HOME` (`~/.local/state`, by default) or `$XDG_DATA_HOME` (`~/.local/share`) are more appropiate places to store those. `$XDG_STATE_HOME` seems especially appropriate for `Backups`, because these backups are (previous) states of an application.

Moreover, spicetify's current behavior can be annoying from a more practical standpoint. `Extracted/` contains tons of uninteresting, minimized `.js` files. These make recursively grepping for strings in  `~/.config`  annoying (near impossible, actually, because a match from one of those files can easily fill 10000+ lines of scrollback, and they nearly always match). Moreover, people like to backup or VC their `~/.config`. I think a lot of people would find including spotify backups and extractions completely undesirable in these contexts (`spicetify/` is 5000x times larger than most directories in my `~/.config`).

I know both of these practical issues can be worked around, but I think spicetify should do the right thing here, rather than to expect users to circumvent its arguably non-standard behavior.

## Alternative proposals

As an alternative, the backups and extracted files could be saved as (compressed) archives. That way, their contents are ignored when performing searches, and they potentially take up less space. (my `~/.config/spicetify/Extracted` compresses really well — >4 times smaller with zstd compression)

rxri:
Spicetify v2 is in currently maintenance mode so your proposal won't be considered by me because I don't even have linux installation rn. Feel free to make a pull request changing it of course.

mirbilal:
Out of this world stuff

caraar12345:
I would advocate for this on macOS as well as Linux -- `~/.local` and `~/.config` are used in much the same way over here too!

rxri:
Resolved by https://github.com/spicetify/cli/releases/tag/v2.38.8

PR #3282, opened 2025-01-26
refactor: change folder location for `Backup` and `Extracted`

UnaTried:
This was requested on issue #3201, I don't know how to test this, so I don't know if it will work. It should in theory though!

P.S: I am not a go-er I just made those changes, did `go build` and there was no ouput, so I assumed it worked, since it was throwing error whenever there was one.

### Linux only version
```
func GetStateFolder(name string) string {
	result, isAvailable := os.LookupEnv("SPICETIFY_STATE")
	defer func() { CheckExistAndCreate(result) }()
		
	if isAvailable && len(result) > 0 {
		return result
	}

	if runtime.GOOS == "linux" {
		parent, isAvailable := os.LookupEnv("XDG_STATE_HOME")

		if !isAvailable || len(parent) == 0 {
			parent = filepath.Join(os.Getenv("HOME"), ".local", "state")
			CheckExistAndCreate(parent)
		}

		result = filepath.Join(parent, "spicetify")
	}
	return result
}

```


rxri (changes requested):
it doesn't return anything when env variable is undefined & os isn't linux tho

UnaTried:
As I said, I do not use go, I just tried to fix that issue using GetUserFolder and GetSpicetifyFolder

rxri:
Then revert your last commit you made?

UnaTried:
Ok, I have reverted my changes @rxri

rxri (approved):

PR #3308, opened 2025-02-11
feat(lyrics-plus): translation Below-mode Individual copy implementation

Lseoksee:
If the lyrics are in Below-mode, the original lyrics or the translated lyrics can now be copy depending on the cursor position.

Additionally, when changing the `musixmatch translation language` in `lyrics plus config`, the page is now designed to re-load, allowing the translation feature to be used without switching to the next song.

And minor bugs were fixed, and unnecessary code was revised

- lyrics copy

https://github.com/user-attachments/assets/62d9a1d9-ba06-42ea-9bc9-2159109f0bbf

- musixmatch translation language change

https://github.com/user-attachments/assets/dda2972c-3a28-4c87-8bbf-8dac628c3f18

### DIFF
--- a/CustomApps/lyrics-plus/OptionsMenu.js
+++ b/CustomApps/lyrics-plus/OptionsMenu.js
@@ -92,8 +92,6 @@ const TranslationMenu = react.memo(({ friendlyLanguage, hasTranslation }) => {
 			none: "None",
 		};
 
-		const savedTranslationDisplay = localStorage.getItem(`${APP_NAME}:visual:translate:display-mode`) || "replace";
-		CONFIG.visual["translate:display-mode"] = savedTranslationDisplay;
 		const translationDisplayOptions = {
 			replace: "Replace original",
 			below: "Below original",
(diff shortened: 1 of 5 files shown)


rxri:
ok. converting to draft for now then

Lseoksee:
When changing the `Musixmatch Translation Language` or `switching lyrics providers`, the lyrics cache is cleared, and the lyrics are reloaded.

rxri:
cool

Lseoksee:
fix duplication active issues between `convert` and `translation`.
If `convert` is enabled, `translation` will be disabled, and vice versa.

@rxri  It's not yet. I think I just need to do one last thing.

Lseoksee:
fix duplication active issues between `convert` and `translation`.
If `convert` is enabled, `translation` will be disabled, and vice versa.

@rxri  It's not yet. I think I just need to do one last thing.

Lseoksee:
for songs in languages that support translation but not conversion (e.g., English), the Convert option is disabled.

@rxri I'll wrap it up here. There are more changes, but I'll include them in the next pull request.

rxri (approved):
well copilot didn't find anything so am good with it too

rxri:
> The 'p1' element is not a valid HTML element. It should be 'p'.
return react.createElement("p1", null, reactChildren);

okay, lmao how no one noticed this

PR #3322, opened 2025-02-20
fix(lyrics-plus): `1.2.57+` Lyrics plus page Issue

### LINKED ISSUE #3320: Lyrics plus appearing halved

- [x] I have checked the FAQ

```Markdown
- Spotify version: 1.2.53.440
- Spicetify version: v2.39.3
```

### 📝 Description

One of the users clarified that it's something on Spotify. But I reinstalled the app. It was fine before. Does anyone know what it is, if they have seen this before?

### 📸 Screenshots

Image

Lseoksee:
fix #3320

### DIFF
--- a/CustomApps/lyrics-plus/style.css
+++ b/CustomApps/lyrics-plus/style.css
@@ -704,6 +704,11 @@ div.lyrics-tabBar-headerItemLink {
 .split .lyrics-versionSelector select {
 	width: 100%;
 }
+/* (1.2.57+) lyrics-plus view issues */
+.hYAxYCXtABXg35wNC2TR.bMaQdOqIiAZtt4cKEhWf.KL8t9WB65UfUEPuTFAhO,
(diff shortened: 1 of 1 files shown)


rxri:
That’s hacky. These elements should be mapped first then we can “modify” them

Lseoksee:
> That’s hacky. These elements should be mapped first then we can “modify” them

Should I update `css-map.json`?

rxri:
> > That’s hacky. These elements should be mapped first then we can “modify” them
> 
> Should I update `css-map.json`?

yes. check first if it was mapped in earlier versions, so you can use the same name for them, but if not, you would need to come up with some name for those

Lseoksee:
There were no identical elements from the previous version, so I updated `css-map.json` with appropriate key values.

rxri (approved):

PR #3327, opened 2025-02-23
Add a check for offline.bnk in the home dir of the user.

Hadock-is-ok:
If you install spotify from other methods than snap or flatpak, your offline.bnk will probably be in your home dir (~/.cache/spotify). This adds a check for that.


rxri:
Line 40 already does that. Why you add redundant code?

Hadock-is-ok:
Sorry, just saw you replied to my issue.

PR #3345, opened 2025-03-07
feat(config): add script to update Spicetify config with Spotify path

wintervoid:
Added a script that finds the installation path of Spotify and prompts the user to update Spicetify's config with the correct path. The intention of the script is to make the instillation on linux easier

### DIFF
--- a/check-path.sh
+++ b/check-path.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+# Author: wintervoid
+# Function to find the path of Spotify
+find_spotify_path() {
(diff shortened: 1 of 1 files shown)


rxri:
Finding the directories on linux is already in the source code in  https://github.com/spicetify/cli/blob/main/src/utils/config.go#L301

wintervoid:
> Finding the directories on linux is already in the source code in https://github.com/spicetify/cli/blob/main/src/utils/config.go#L301

Sorry i had no idea that there was already a feature for that lol

PR #3352, opened 2025-03-17
fix(spicetify-wrapper): remove unnecessary scroll check on DOM changes

sanoojes:
I think this solution is much better and only need class name updates (if needed) in future

This solution removes the unnecessary Observer and replaces it with a style sheet that is only applied once to the dom


sanoojes:
@kyrie25 @Cali
Can you check this pr out ? 
I can't figure out what's wrong with the pr

Issue #3356, opened 2025-03-23
Spicetify breaks on multi user systems sharing the same spotify installation (/opt/spotify/Apps)


zjeffer:
### 🔍 Have you checked Spicetify.app page for your issue?

- [x] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
- Spotify version: `1.2.59.514`
- Spicetify version: `2.39.5`
- Arch Linux
- Hyprland
- Spotify and spicetify-cli were both installed through the AUR.
```

### 📝 Description

I have 2 users: one for work and one personal user. My work user is also in my personal user's group, so it has access to all group-accessible files.
I have followed the wiki's instructions to set the correct permissons in `/opt/spotify/Apps`, but these seem to get overwritten to `700` by spicetify at some point? This causes spicetify commands to not work for user 1 if I ran them before on user 2. Spotify also launches with a blank screen for the other user, I assume because it doesn't have access to the Apps folder. 

How do other people manage this kind of setup? I'd like to be able to share the same config between both users without having to run any spicetify commands every time I login to the other user.

(will add some more details to this issue soon, need to login to my other user first)

rxri:
Please install spotify per-user then, and not system-wide. This is not something we should manage in our code.

zjeffer:
Can't spicetify manage it by not changing the permissions of /opt/spotify/Apps? Why does it do that?

rxri:
you can chmod/change perms manually to `770` or something. this is kinda something that shouldn't be by default bc it can do another sort of issues. One spicetify installation for multiple users is not supported.

zjeffer:
That indeed works. I don't see why spicetify currently sets it 700 when applying, though? Why not 750, so that it can at least be read by other users of the same group?

I'd rather not manage Spotify on a per-user basis, I like doing everything through my system package manager.

rxri:
> That indeed works. I don't see why spicetify currently sets it 700 when applying, though? Why not 750, so that it can at least be read by other users of the same group?

To be honest, I don't know. khanhas must have made it this way for some reason or another.
https://github.com/spicetify/cli/blob/main/src/utils/utils.go#L90

rxri:
we could change it to 750 ig?

zjeffer:
Do you mind if I make a PR where I change all these to 750?

rxri:
feel free

[...]

PR #3357, opened 2025-03-23
Use 770 permissions instead of 700, fixes multi user installs #3356

zjeffer:
Will do more testing first


rxri:
did you test it and is it ready for review or...?

zjeffer:
It doesn't fully work, I need to investigate why ~/.local/state/spicetify is still being created with 700 permissions. I haven't had time yet but might soon.

zjeffer:
Just tested it again with a clean slate, by:
* Removing `/opt/spotify` and `/opt/spicetify-cli`
* Uninstalling `spotify` and `spicetify-cli` using my package manager
* Reinstalling spotify, chmodding `/opt/spotify` and `/opt/spotify/Apps` with a+rw like it says in the wiki
* Rebuilding spicetify-cli from my branch with `go build -o spicetify`

Running `./spicetify backup` from the repo creates a ~/.local/state/spicetify/ folder with 700 permissions. Deleting the folder and trying it again now results in a folder with 750 permissions. Not sure what happened the first time, maybe I saw it wrong. Tried it again a couple of times from a clean slate and it now keeps creating the folder with 750 permissions.

Now calling `./spicetify restore` makes /opt/spotify/Apps have the same 750 permissions, because it copies the state folder of course.

rxri:
why does it create them with 750 tho instead of 770

PikachuGabe:
from what I can find, it is caused by https://stackoverflow.com/a/29895320 .

PR #3378, opened 2025-04-22
feat: add getResizeHost

Donaldino7712:


rxri:
Sorry, but no. We do not want to maintain resize host in wrapper itself.

PR #3468, opened 2025-07-30
Fix error with Spicetify.ReactJSX being undefined when laoding autoSkipExplicit.js

highjeans:
There should be a better way to do this but this is sort of a "hot fix".

### DIFF
--- a/Extensions/autoSkipExplicit.js
+++ b/Extensions/autoSkipExplicit.js
@@ -5,7 +5,7 @@
 /// <reference path="../globals.d.ts" />
 
 (function ChristianSpotify() {
-	if (!Spicetify.LocalStorage) {
+	if (!Spicetify.LocalStorage || !Spicetify.ReactJSX?.jsx) {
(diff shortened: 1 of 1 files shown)


rxri:
but `Spicetify.ReactJSX` is not used anywhere in this code.

highjeans:
It's used internally, I found it from the error message.
`new Spicetify.Menu.Item()` calls `new Spicetify.ContextMenuV2.Item` which then has the line `this._element = Spicetify.ReactJSX.jsx(() => {` that ends up being undefined

rxri:
> It's used internally, I found it from the error message. `new Spicetify.Menu.Item()` calls `new Spicetify.ContextMenuV2.Item` which then has the line `this._element = Spicetify.ReactJSX.jsx(() => {` that ends up being undefined

Then `Spicetify.Events.webpackLoaded` exists...

highjeans:
honestly I didn't take much of a look at the code lol, hence the description of the pr: "there should be a better way to do this"

PR #3516, opened 2025-09-10
fix(globals): add isRight property to Button class

D3SOX:
Seemed like an oversight to me but I'd like to hear your feedback on this. I have not checked other classes for outdated types.

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -1814,11 +1814,12 @@ declare namespace Spicetify {
 	 */
 	namespace Topbar {
 		class Button {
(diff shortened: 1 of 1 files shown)


rxri:
> who didn't even understand that this is not a new feature

It did understand that it's only types update but only updating types

D3SOX:
> It did understand that it's only types update but only updating types

Yeah but it literally said

> New Features
>    Topbar buttons can now be aligned to the right

but that was already possible before.  🤷

I was just pointing it out because in my opinion this bot made the PR way less readable

PR #3521, opened 2025-09-17
fix: update spicetifyWrapper.js

EnderPoint07:
fixed URI.fromString and URI.from functions 

new update of spotify had them weirdly nested in objects

this was breaking [spicetify-visualizer](https://github.com/Konsl/spicetify-visualizer/tree/main) app 

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of URI handling to reduce errors when parsing or opening Spotify links.
  - Increased compatibility across app versions by making URI detection more robust to internal structure changes.
  - Mitigated edge-case failures that could previously break features relying on URIs.


EnderPoint07:
i just replaced the old searching criteria with the correct one 😭 it aint that deep

EnderPoint07:
the type error thing is find ig even it the docs it is said that not inputting a string to the func will give type err

rxri:
This is already fixed in the repo

Issue #3592, opened 2025-11-23
[Lyrics-Plus] [Critical] After an update, spotify instantly crash after opening lyrics plus


Fortress937:
### 🔍 Have you checked Spicetify.app page for your issue?

- [x] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
Spotify for Windows (64 bit)
1.2.76.298.g2316a870
Spicetify v2.42.1
```

### 📝 Description

After receiving an update from spicetify, I update it, after an update, opening lyrics plus cause it to crash.

If no one experiencing this issues, this issues is likely at my end and you can close my issues, if someone experienced the same, consider fixing it.

You can still use spotify, but will crash when opening lyrics plus

### 📸 Screenshots

https://github.com/user-attachments/assets/4db4152c-0fa4-4b47-8dd0-b8f03d763b01

Fortress937:
<img width="573" height="687" alt="Image" src="https://github.com/user-attachments/assets/1428b6ae-ce97-447c-9090-720c2129187e" />

Ok that literally affect everyone, and reinstalling lyrics plus probably won't help (Yeah i did and it didn't work. )

rxri:
Cannot reproduce on `1.2.76.298`. Reinstall everything (as of uninstall, clean up and then install again)

Fortress937:
> Cannot reproduce on `1.2.76.298`. Reinstall everything (as of uninstall, clean up and then install again)

Confirm this also break other spotify element, well thanks for help (Much aware of this update)

Fortress937:
> Cannot reproduce on `1.2.76.298`. Reinstall everything (as of uninstall, clean up and then install again)

<img width="740" height="682" alt="Image" src="https://github.com/user-attachments/assets/eb32870a-7713-4e2f-8190-ebdf02a0471b" />

It work for a second but after i enabling Musixmatch provider, it crashes, I think it has something to do with the provider Musixmatch.

PR #3596, opened 2025-11-26
fix(lyrics-plus): fix issue with non-standard translations from Musixmatch providers

Lseoksee:
In the case of the non-standard translation of the Musixmatch provider, the `DisplayNames.of()` function would return an exception instead of `null` or `undefined`, causing Spotify to crash.

Therefore, we modified this to handle exceptions using a try-catch statement.

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling in the translation menu so label resolution now uses a safe lookup path with an explicit fallback. Labels will reliably display and will fall back to an uppercase language code when resolution fails, preventing UI errors and missing labels.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/CustomApps/lyrics-plus/OptionsMenu.js
+++ b/CustomApps/lyrics-plus/OptionsMenu.js
@@ -129,7 +129,7 @@ const TranslationMenu = react.memo(({ friendlyLanguage, hasTranslation, musixmat
 			const musixmatchOptions = availableMusixmatchLanguages.reduce((acc, code) => {
 				let label = "";
 				try {
-					label = musixmatchDisplay.of(code) || code.toUpperCase();
+					label = musixmatchDisplay.of(code) ?? code.toUpperCase();
 				} catch (e) {
 					label = code.toUpperCase();
 				}


Lseoksee:
fix issue #3592.

CustomApps/lyrics-plus/OptionsMenu.js
```diff
@@ -127,7 +127,12 @@ const TranslationMenu = react.memo(({ friendlyLanguage, hasTranslation, musixmat
 
 		if (availableMusixmatchLanguages.length) {
 			const musixmatchOptions = availableMusixmatchLanguages.reduce((acc, code) => {
-				const label = musixmatchDisplay.of(code) || code.toUpperCase();
+				let label = "";
+				try {
```
rxri, line 131: what's the point of try {} catch {} if you use OR operator that does the same thing?

Lseoksee: Referring to the documentation at 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of#return_value

If the function argument isn't a language code, it seems to raise a `RangeError` exception. This means it doesn't return a value like null or undefined, so it can't be processed with the OR operator.

However, if it's a language code but not included in `DisplayNames`, it returns undefined .

rxri: Use `??` then and I can approve this.

Lseoksee: Done edit.

PR #3663, opened 2025-12-18
Refactor album and artist track fetching

marmik-7:
- Simplify fetchAlbumTracks to use a cleaner pagination loop and consistent filtering.
- Refactor fetchArtistLikedTracks for clearer mapping and filtering of playable tracks.
- Update fetchQueue to merge current, next, and queued tracks without duplicates.


rxri:
Another pointless pull request. It uses fisher-yates alghorithm. Duplicates are SUPPOSED to happen

PR #3673, opened 2025-12-31
Better offline.bnk detection (linux)

b33-stinger:
`devtools.go` didn't check whether the path actually has `offline.bnk` before using it as the `filePath`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline cache detection for Snap and Flatpak installations on Linux, ensuring the application correctly locates cached data in the appropriate package manager directories.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/cmd/devtools.go
+++ b/src/cmd/devtools.go
@@ -28,18 +28,21 @@ func EnableDevTools() {
 		{
 			homePath := os.Getenv("HOME")
 			snapSpotifyHome := homePath + "/snap/spotify/common"
-			if _, err := os.Stat(snapSpotifyHome); err == nil {
-				homePath = snapSpotifyHome
-			}
-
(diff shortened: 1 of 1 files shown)


rxri:
Flatpak version stores `offline.bnk` in `<flatpak home>/config/spotify/prefs`. The else statement is for normal spotify installs

b33-stinger:
> Flatpak version stores `offline.bnk` in `<flatpak home>/config/spotify/prefs`. The else statement is for normal spotify installs

i had a problem that i unknowingly had a spotify folder from flatpak but i didn't use flatpak spotify, so it didn't have the `offline.bnk` file and it would always tell be "couldn't find offline.bnk" (because it detected flatpak spotify and skipped my "normal" path). My pr just makes sure `offline.bnk` exists in the folder before using it.
Not something worth patching?

PR #3692, opened 2026-02-05
fix(popupLyrics): fix lyrics not syncing when minimized

Nordara:
use setTimeout fallback when document is hidden to keep tick running

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved popup lyrics reliability by adding a fallback timing mechanism so updates continue when the browser tab is hidden or not in focus.
  * Preserved the existing smooth update behavior when the tab is visible while ensuring consistent background updates to avoid missed lyric progress.


rxri:
`popupLyrics` utilize webworkers when spotify is minimized. I just tried and it works just fine, there's no need for this timeout. coderabbit also pointed it out

PR #3701, opened 2026-02-17
fix(platform): defer PlatformAPI resolution to prevent Linux segfault

### LINKED ISSUE #3695: After applying spicetify the Spotify segfaults (`1.2.82.428`)

- [x] I have checked the FAQ

```Markdown
- Spotify version: 1.2.82.428.g0ac8be2b
- Spicetify version: 2.42.8
```

### 📝 Description

updated Spotify to latest version and ran  `spicetify backup apply` .  spotify no longer opens. It closes on by itself

### 📸 Screenshots

https://github.com/user-attachments/assets/d2f65f80-8b8a-41b8-a7bd-4ca299c3481b

enXov:
## Summary

Fixes #3695 — Spotify 1.2.82 on Linux segfaults on launch after `spicetify backup apply`.

## Root Cause

Spicetify eagerly calls all `_platform.get*()` native getter functions at startup in `waitForPlatform()`. On Spotify 1.2.82 for Linux, one or more of these native getters triggers a SIGSEGV (segmentation fault) in Chromium's native layer. Since this is a native crash, JavaScript `try-catch` cannot intercept it.

## Fix

Convert all PlatformAPI getter resolution from **eager** (called at startup) to **lazy** (`Object.defineProperty` getters that resolve on first access):

- Native code is never called at startup — zero getter invocations during init
- Each API self-caches its resolved value on first access
- Failed resolutions return `undefined` instead of crashing
- Backward-compatible: on Windows/macOS, APIs resolve identically on first access

### [jsHelper/spicetifyWrapper.js](cci:7://file:///home/enxov/OFK/Projects/Github-PR/cli/jsHelper/spicetifyWrapper.js:0:0-0:0)
- `waitForPlatform()`: Replace eager `_platform[key]()` calls with lazy `Object.defineProperty` getters
- `addMissingPlatformAPIs()`: Replace eager `Registry.resolve()` calls with lazy getters
- Add defensive error handling for version parsing and Registry iteration

### [jsHelper/homeConfig.js](cci:7://file:///home/enxov/OFK/Projects/Github-PR/cli/jsHelper/homeConfig.js:0:0-0:0)
- Add null check for `Spicetify.Platform.History` before accessing `.location`

### [jsHelper/expFeatures.js](cci:7://file:///home/enxov/OFK/Projects/Github-PR/cli/jsHelper/expFeatures.js:0:0-0:0)
- Add optional chaining for `RemoteConfigDebugAPI?.getProperties()`

### [src/apply/apply.go](cci:7://file:///home/enxov/OFK/Projects/Github-PR/cli/src/apply/apply.go:0:0-0:0)
- [insertExpFeatures](cci:1://file:///home/enxov/OFK/Projects/Github-PR/cli/src/apply/apply.go:460:0-489:1): Wrap [expFeatureOverride](cci:1://file:///home/enxov/OFK/Projects/Github-PR/cli/jsHelper/expFeatures.js:17:1-42:3) hook with `typeof` check + try-catch
- [insertExpFeatures](cci:1://file:///home/enxov/OFK/Projects/Github-PR/cli/src/apply/apply.go:460:0-489:1): Wrap `RemoteConfigResolver` and `createInternalMap` captures with try-catch
- [insertHomeConfig](cci:1://file:///home/enxov/OFK/Projects/Github-PR/cli/src/apply/apply.go:408:0-431:1): Guard `SpicetifyHomeConfig.arrange` with `typeof` check + fallback

## Testing

- 15 unit tests added ([jsHelper/__tests__/platformAPI.test.js](cci:7://file:///home/enxov/OFK/Projects/Github-PR/cli/jsHelper/__tests__/platformAPI.test.js:0:0-0:0)), all passing
- Manually verified on Linux with Spotify 1.2.82.428 — Spotify opens, music plays, Marketplace loads, no segfault

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests validating lazy Platform API resolution, caching, error handling, enumeration, and registry behavior.

* **Bug Fixes**
  * Strengthened runtime guards and error handling to prevent crashes when Platform pieces are missing.
  * Home initialization now safely skips when required components are absent.

* **New Features**
  * Introduced lazy, on-first-access loading and caching for Platform and registry-provided APIs.

### DIFF
--- a/src/apply/apply.go
+++ b/src/apply/apply.go
@@ -482,7 +482,7 @@ func insertExpFeatures(jsPath string, flags Flag) {
 			&content,
 			`(([\w$.]+\.fromJSON)\(\w+\)+;)(return ?[\w{}().,]+[\w$]+\.Provider,)(\{value:\{localConfiguration)`,
 			func(submatches ...string) string {
-				return fmt.Sprintf("%stry{Spicetify.createInternalMap=%s}catch(_e){};%stry{Spicetify.RemoteConfigResolver=%s}}catch(_e){};{value:{localConfiguration", submatches[1], submatches[2], submatches[3], submatches[4])
+				return fmt.Sprintf("%stry{Spicetify.createInternalMap=%s}catch(_e){};%sSpicetify.RemoteConfigResolver=%s", submatches[1], submatches[2], submatches[3], submatches[4])
 			})
 
 		return content


rxri:
I appreciate the want of the fix and the explanation of the issue, however:
- This is a total AI slop
- A lot of extensions WAIT for the stuff to appear in the `Spicetify.Platform` this only prevents them from working properly and waiting indefinitely

I'm not really happy about it and am not even considering it to be merged

enXov:
Thanks for the review @rxri. I understand the concern about extensions, so let me clarify how the lazy getters actually work:

Extensions are NOT affected. When an extension accesses Spicetify.Platform.PlayerAPI, the lazy getter fires immediately, calls the native getter, caches the result, and returns it — all in the same access. From the extension's perspective, the property is there and works normally. There's no waiting or polling needed.

The only difference from the current behavior is when the native call happens:

Before: All ~20 native getters called eagerly at startup → one crashes → whole app segfaults
After: Each native getter called on first access → only the broken one fails → returns undefined → app continues
On Windows/macOS, the behavior is functionally identical — every API resolves on first access. On Linux 1.2.82, without this fix, Spotify crashes immediately after spicetify backup apply. The current workaround on the issue thread (#3695) is disabling home_config and experimental_features, which breaks Marketplace. This fix keeps everything enabled.

[Screencast_20260217_235859.webm](https://github.com/user-attachments/assets/9f2109fe-70b0-428e-9330-0573be28ea2f)

However, if you dont like the code you can reject it 😭 ❤️

rxri:
What's the point of this if only one getter will always fail? Let's just find the one that doesn't work and not call it at all lol. This is literally over-engineered and I can tell it's AI because they like to do this even for simple issues

enXov:
Thanks for the suggestion but the real problem is on the spotify sides; what i meant, the crash happening on the spotify native c++ codes, so the js cannot catch the errors unfortuanly. What is that meaning is need extremely debug, so thats the basic solution. And yeah im using AI XD 

but yeah, in general your right, if you want to me contunie i will contunie to debug

rxri:
I am aware this is spotify's fault. They broke it. I said since the beginning that this is not our fault and this just proves it further.

> so thats the basic solution

It's over engineered for no reason. The AI should not be used for such an issue anyway. The `createInternalMap` and `expFeatureOverride` do not exist on newer Spotify versions because there's an actual PlatformAPI for it; `RemoteConfigDebugAPI`. It's only keep for compatibility. If disabling home config & exp features make spotify to not crash - it's probably something around `RemoteConfigDebugAPI` & something else

enXov:
yes, but if the home config and exp features is disabled then no marketplace and the spicetify without marketplace is nothing(this time include extensions)

so the as i said, binary research is needed, and i skip that one, dont want to spend time on it, and use the working one since i dont need every api on spotify, so you can close the pr ❤️

PR #3705, opened 2026-02-20
fix: skip ExclusiveModeAPI resolution to prevent Linux segfault

romdotdog:
Needs platform check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of platform API initialization by adding an OS-specific guard to avoid resolving an incompatible API on Linux, while preserving existing resolution behavior and logging on other platforms.
  * No changes to public/exported interfaces.

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -326,7 +326,7 @@ window.Spicetify = {
 		setTimeout(addMissingPlatformAPIs, 50);
 		return;
 	}
-	const os = Spicetify.Platform?.operatingSystem;
+	const os = Spicetify.Platform.operatingSystem;
 	const version = Spicetify.Platform.version.split(".").map((i) => Number.parseInt(i, 10));
 	if (version[0] === 1 && version[1] === 2 && version[2] < 38) return;
 


rxri:
-.- lossless related feature doing bugs niceu, we should get `operatingSystem` from `Spicetify.Platform` and check if it's Linux and only disable it for this platform

jsHelper/spicetifyWrapper.js
```diff
@@ -326,13 +326,14 @@ window.Spicetify = {
...
 	}
+	const os = Spicetify.Platform?.operatingSystem;
 	const version = Spicetify.Platform.version.split(".").map((i) => Number.parseInt(i, 10));
 	if (version[0] === 1 && version[1] === 2 && version[2] < 38) return;
 
 	for (const [key, _] of Spicetify.Platform.Registry._map.entries()) {
 		if (typeof key?.description !== "string" || !key?.description.endsWith("API")) continue;
 		const symbolName = key.description;
-		if (symbolName === "ExclusiveModeAPI") continue;
+		if (symbolName === "ExclusiveModeAPI" && os === "Linux") continue;
```
rxri, line 336: it's `Linux` for 100%? and not for example `linux`?

romdotdog: I `console.log`ed and tested it

jsHelper/spicetifyWrapper.js
```diff
@@ -326,12 +326,14 @@ window.Spicetify = {
 		setTimeout(addMissingPlatformAPIs, 50);
 		return;
 	}
+	const os = Spicetify.Platform?.operatingSystem;
```
rxri:
```suggestion
	const os = Spicetify.Platform.operatingSystem;
```

rxri:
ty

PR #3707, opened 2026-02-23
fix: resolve resource leaks and add shared HTTP client

iso2kx:
## Summary
- Fix `defer` inside loop in `Unzip()` and `Copy()` — file descriptors accumulated until the loop completed, potentially exhausting OS limits on large archives. Extracted helper functions so `defer` runs per iteration.
- Fix 3 HTTP response body leaks (`vcs.go`, `watcher.go`, `preprocess.go`) by adding missing `defer resp.Body.Close()`.
- Add shared `utils.HTTPClient` with 30-second timeout. The default `http.Get` uses a client with no timeout and can hang indefinitely.
- Optimize `Watch()` and `WatchRecursive()` to compare file mtime/size instead of reading entire file contents every 200ms.

## Files Changed
- `src/utils/utils.go` — Extract `extractZipEntry()` and `copyOneFile()` helpers
- `src/utils/http.go` — New shared HTTP client
- `src/utils/vcs.go` — Body leak fix + use `HTTPClient`
- `src/utils/watcher.go` — Body leak fix + mtime-based polling
- `src/cmd/update.go` — Use `HTTPClient`
- `src/preprocess/preprocess.go` — Body leak fix + use `HTTPClient`

## Test plan
- [ ] `spicetify apply` works correctly
- [ ] `spicetify watch` detects file changes
- [ ] `spicetify update` downloads successfully
- [ ] No goroutine/fd leaks under repeated apply cycles

## Release Notes

* **Improvements**
  * HTTP requests across the application now use a unified client with a 30-second timeout, preventing indefinite hangs during update downloads and asset fetches.
  * File change detection has been optimized to compare file metadata (modification time and size) instead of reading full file contents, significantly reducing resource consumption and improving overall performance.

### DIFF
--- a/src/cmd/update.go
+++ b/src/cmd/update.go
@@ -55,7 +55,7 @@ func Update(currentVersion string) bool {
 	}
 	defer out.Close()
 
-	resp2, err := http.Get(assetURL)
+	resp2, err := utils.HTTPClient.Get(assetURL)
 	if err != nil {
 		spinner.Fail("Failed to download Spicetify")
 		utils.Fatal(err)
--- a/src/preprocess/preprocess.go
+++ b/src/preprocess/preprocess.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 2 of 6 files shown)


rxri:
AI slop. Literally none of this is an issue

PR #3708, opened 2026-02-23
perf: optimize preprocessing with regex cache, batch CSS, and parallelism

iso2kx:
## Summary
- Add `sync.Map`-based regex compilation cache — the `Replace`, `ReplaceOnce`, `FindMatch`, `FindSymbol`, and `SeekToCloseParen` functions were recompiling identical regex patterns on every call across thousands of file patches.
- Hoist static regexes to package level in `color.go` (`hexColorRegex`, `xrdbRegex`) and `preprocess.go` (`uriRegex`, `buildRegex`).
- Replace per-entry CSS translation map loop (`for k, v := range cssTranslationMap { utils.Replace(...) }`) with `strings.NewReplacer` — this uses Aho-Corasick internally, turning N full-file regex scans into a single efficient pass per file.
- Replace `filepath.Walk` with `filepath.WalkDir` to avoid redundant `os.Lstat` syscalls.
- Parallelize file patching with a bounded worker pool (`runtime.NumCPU()` goroutines) since each file is independently read, modified, and written back.

## Files Changed
- `src/utils/utils.go` — Add regex cache, update 7 functions to use it
- `src/utils/color.go` — Hoist regexes to package level
- `src/preprocess/preprocess.go` — Hoist regexes, `NewReplacer`, `WalkDir`, parallel loop

## Test plan
- [ ] `spicetify apply` produces identical output to before
- [ ] CSS class translations still work correctly
- [ ] `spicetify apply` is noticeably faster on multi-core machines
- [ ] No race conditions (run with `-race` flag)

## Summary by CodeRabbit

* **Refactor**
  * Optimized file processing performance through concurrent worker pattern and centralized regex caching, improving overall application responsiveness.
  * Streamlined CSS patching application for efficiency.
  * Enhanced file system traversal mechanism.

### DIFF
--- a/src/preprocess/preprocess.go
+++ b/src/preprocess/preprocess.go
@@ -13,9 +13,15 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/pterm/pterm"
 	"github.com/spicetify/cli/src/utils"
+)
+
+var (
+	uriRegex   = regexp.MustCompile(`(?:class ([\w$_]+)\{constructor|([\w$_]+)=function\(\)\{function ?[\w$_]+)\([\w$.,={}]+\)\{[\w !?:=.,>&(){}[\];]*this\.hasBase62Id`)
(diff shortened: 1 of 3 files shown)


rxri:
AI slop. regex must compile is not that expensive

PR #3771, opened 2026-03-30
feat(loopyLoop): complete rehaul

3xyy:
Complete rehaul of loopyLoop. Now persists across sessions for a song, user can repeat the song with Spotify if they want to loop instead of the plugin doing it, added ability to skip sections, new button in toolbar, can adjust section boundaries and start/end boundaries by right clicking. If Spotify is playing from phone but you have spicetify open on your computer it will follow the []{} rules set by the extension

image

## Summary by CodeRabbit

* **New Features**
  * Per-song start/end markers and persistent per-track skip zones with automatic reload on song change
  * Toolbar button opens a fixed-position context menu for creating, editing, moving and removing markers/zones
  * Pending marker visuals and a “Move” submenu to nudge marker/zone positions

* **Improvements**
  * Reworked playback behavior: restart/previous handling, boundary enforcement, auto-advance and skip-zone seeking
  * More reliable progress-bar marker rendering, selection and right-click support

### DIFF
--- a/Extensions/loopyLoop.js
+++ b/Extensions/loopyLoop.js
@@ -564,10 +564,18 @@
 		true
 	); // capture phase
 
-	// Load state for the currently playing song on startup
-	loadState();
-	drawOnBar();
-	drawSkipMarkers();
+	// Load state for the currently playing song on startup.
+	// Retry until the player has track data (uri may be null immediately after init).
+	function tryLoadInitialState(attemptsLeft) {
+		if (Spicetify.Player.data?.item?.uri) {
+			loadState();
+			drawOnBar();
+			drawSkipMarkers();
+		} else if (attemptsLeft > 0) {
+			setTimeout(() => tryLoadInitialState(attemptsLeft - 1), 200);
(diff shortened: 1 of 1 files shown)


3xyy:
@coderabbitai review

rxri (approved):

rxri:
It still would be nice for this extension to use React in the future

NitrousWarrior:
Unfortunately, the loop doesn't work. It just skips to the next song without replaying the part you want it to loop. Even if you enable the repeat button, it still skips to the next song. You can technically do it with the new skip feature, but it is a bit annoying to do so. I prefer the old loop system more.

3xyy:
@NitrousWarrior > Unfortunately, the loop doesn't work. It just skips to the next song without replaying the part you want it to loop. Even if you enable the repeat button, it still skips to the next song. You can technically do it with the new skip feature, but it is a bit annoying to do so. I prefer the old loop system more.

#3833 - good catch, I swear it worked before.

PR #3786, opened 2026-04-07
fix(devtools): fail safely when app-developer marker is missing in offline.bnk

nghiphaam:
strings.Index returns -1 if nothing is found but the old code just used it as an offset directly, so if spotify ever changes the app-developer marker it would write at a completely wrong position and corrupt offline.bnk.

extracted the lookup into its own function that returns an error instead, and hooked up a proper exit in EnableDevTools. also added some tests while i was at it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in developer tools functionality with explicit error reporting when required markers are missing.

* **Tests**
  * Added unit test coverage for developer marker detection in offline bank files, including scenarios with missing and multiple marker occurrences.


rxri:
what? but they never removed it so what's the point of this?

nghiphaam:
> what? but they never removed it so what's the point of this?

Actually there's a small mistake. it's not byte 13/14, it's byte 13 and 14 only if both markers are missing. but LastIndex on the same string would also return -1, so both writes land at -1 + 14 = 13 and -1 + 15 = 14. so that part is correct.

rxri:
remove the tests since we don't have them

nghiphaam:
> remove the tests since we don't have them

kk i will make a commit in free time

nghiphaam:
i did

rxri (approved):

PR #3817, opened 2026-04-17
chore(lyrics-plus): correct misspelled variable name

nghiphaam:
Cache lookup for translated lyrics was using an incorrectly named variable, causing the cache check to always miss and re-process lyrics unnecessarily on every track load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected a typo in the lyrics caching logic with no impact to functionality or user experience.


rxri (approved):

rxri:
this is just a variable name?
I mean sure will merge it but it has nothing to do with cache xD

PR #3839, opened 2026-05-01
feat(service): add auto-apply on Spotify updates

manolopro3333:
This change adds an automatic self-healing service for Spicetify. It runs in the background, checks for new Spicetify releases on startup and every 6 hours, and reapplies the patch when needed. The installer now enables the service automatically, and `spicetify restore` disables it so user intent is respected.

## Release Notes

* **New Features**
  * Introduced background service for automatic Spicetify maintenance and repair
  * Added `spicetify service` command with install, enable, disable, and status subcommands
  * Automatic update checks on system startup and at configurable intervals
  * Service auto-update configuration options now available

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -210,4 +210,11 @@ else {
   Invoke-WebRequest @Parameters | Invoke-Expression
 }
 #endregion Marketplace
+
+#region Service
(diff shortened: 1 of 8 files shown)


rxri:
Spicetify was never designed to run in background, and also it's out of scope of the "CLI" app. Sorry

PR #3844, opened 2026-05-18
fix(loopyLoop): update progress bar selectors to use data-testid

frane0:
Spotify's recent UI updates removed the legacy .playback-bar and .playback-progressbar-container classes, causing the extension to fail silently. This commit replaces those class selectors with the persistent data-testid="playback-progressbar" attribute to restore context menu and UI injection functionality.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of playback progress bar detection to ensure consistent interaction handling across different UI states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/spicetify/cli/pull/3844?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Extensions/loopyLoop.js
+++ b/Extensions/loopyLoop.js
@@ -7,8 +7,7 @@
 /// <reference path="../globals.d.ts" />
 
 (function LoopyLoop() {
-	const playbackBar = document.querySelector(".playback-bar");
-	const progressContainer = playbackBar?.querySelector(".playback-progressbar-container");
+	const progressContainer = document.querySelector('[data-testid="playback-progressbar"]');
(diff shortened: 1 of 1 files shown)


WiLuX-Source:
I can confirm this fixes the issue for plugin loopyloop. Is it a long term solution though?

frane0:
> I can confirm this fixes the issue for plugin loopyloop. Is it a long term solution though?

To be honest, I don't know. Because Spotify can change it again anytime. If they eventually remove the data-testid, we can just find the new selector and patch it again.

rxri:
These classes were mapped by us and were not "legacy"

PR #3880, opened 2026-07-04
fix: support rspack builds (Spotify 1.2.93+)

iso2kx:
## Problem

Spotify 1.2.93 migrated its client bundle from **webpack to rspack**. This breaks Spicetify in several ways on current Spotify, with no workaround from config:

- **Extensions & context menus break** — `Spicetify.React` and `Spicetify.URI` resolve to `undefined` ("Something went wrong" on right-click).
- **Custom apps don't mount** — Marketplace (and other custom apps) render a blank page with `useNavigateStable must be used within a StableUseNavigateProvider`, and `apply` prints "Spotify version mismatch".

## Root causes & fixes

**Wrapper (`jsHelper/spicetifyWrapper.js`)**
- rspack renames the chunk global to `rspackChunkclient_web`; accept either global.
- rspack exposes proxy-trap module exports that return an object for any property and throw on `toString`, matching every symbol heuristic. Filter them out of the module cache.
- The URI chunk lookup accepted any truthy `PLAYLIST_V2` (a proxy-trap satisfied it); require a string.

**Custom apps (`src/apply/apply.go`, `src/cmd/apply.go`)**
- Routing moved out of `xpui.js` (which no longer exists on rspack) into `xpui-modules.js`; run `insertCustomApp` against that file too. The existing regexes already match rspack's output — only the target file was wrong.
- Generated custom-app chunks registered with `webpackChunkclient_web`, which the rspack runtime never reads, causing `ChunkLoadError`. Register with `rspackChunkclient_web` when present, falling back to the webpack global for older Spotify.

## Testing

Built from this branch and applied against **Spotify 1.2.93.667** (rspack). After applying:
- Extensions load and context menus work again (`Spicetify.React`/`URI` resolve).
- Marketplace mounts and renders its catalog.
- Verified older-Spotify paths are preserved (webpack global fallback, `xpui.js` still targeted when present).

Backwards compatible: all changes fall back to the previous webpack behaviour when the rspack globals/files are absent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with newer Spotify builds, reducing issues when loading or refreshing custom apps.
  * Fixed module detection so app routing and URI handling work more reliably across different build types.
* **New Features**
  * Added broader support for modern build formats, helping the app continue working on updated releases without manual changes.

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -519,13 +519,30 @@ const fnStr = (f) => {
 };
 
 (async function hotloadWebpackModules() {
-	while (!window?.webpackChunkclient_web) {
+	// Spotify 1.2.93+ builds with rspack, which renames the chunk global.
+	while (!(window?.webpackChunkclient_web || window?.rspackChunkclient_web)) {
 		await new Promise((r) => setTimeout(r, 50));
 	}
+	const chunkGlobal = window.webpackChunkclient_web || window.rspackChunkclient_web;
 
 	// Force all webpack modules to load
-	const require = webpackChunkclient_web.push([[Symbol()], {}, (re) => re]);
+	const require = chunkGlobal.push([[Symbol()], {}, (re) => re]);
 	while (!require.m) await new Promise((r) => setTimeout(r, 50));
+
+	// rspack exposes proxy-trap module exports that return an object for any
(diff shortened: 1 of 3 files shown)


rxri:
I'm not going to merge half-working PR. Thanks for trying though and I was well aware of what's broken

PR #3882, opened 2026-07-04
fix(installer): keep spicetify in process PATH

### LINKED ISSUE #3871: Marketplace installer enters recursive install loop after CLI installer on Windows

- [x] I have checked the FAQ

```Markdown
### Environment

* OS: Windows
* Shell: PowerShell
* Spotify version: `1.2.92.147`
* Spicetify version: `2.43.2`
```

### Summary

When installing Spicetify on Windows using the official PowerShell installer, the CLI installation completes successfully. However, if the user chooses to install Spicetify Marketplace immediately afterward, the Marketplace installer may fail to detect the newly installed `spicetify` command.

Because of that, the Marketplace installer assumes Spicetify is not installed and starts the main CLI installer again. After the CLI installer finishes, it asks again whether to install Marketplace. Choosing `Y` repeats the same process, causing an infinite install loop.

### Command used

```powershell
iwr -useb https://raw.githubusercontent.com/spicetify/cli/main/install.ps1 | iex
```

### What happened

The CLI installer completed successfully:

```text
Installing spicetify...
Fetching the latest spicetify version... > OK
Downloading spicetify v2.43.2... > OK
Extracting spicetify... > OK
Making spicetify available in the PATH... > OK
spicetify was successfully installed!

Run spicetify -h to get started
```

Then the installer asked:

```text
Do you also want to install Spicetify Marketplace?
[Y] Yes  [N] No  [?] Help (default is "Y"):
```

After selecting `Y`, the Marketplace installer started, but it could not find `spicetify`:

```text
Starting the spicetify Marketplace installation script..
Setting up...
Spicetify not found.
Installing it for you...
```

It then launched the CLI installer again. After that finished, the same Marketplace prompt appeared again. Pressing `Y` repeated the same loop.

### Expected behavior

After the CLI installer successfully installs Spicetify and adds it to PATH, the Marketplace installer should either:

1. use the installed `spicetify.exe` directly from the known install path, or
2. refresh/update the current PowerShell session PATH before checking `Get-Command spicetify`, or
3. avoid re-running the CLI installer if Marketplace was launched from the CLI installer itself.

### Actual behavior

The Marketplace installer uses a command check, but in the same PowerShell session the newly added PATH may not be available yet. As a result, it incorrectly reports:

```text
Spicetify not found.
Installing it for you...
```

This causes the main CLI installer to run again, which again asks to install Marketplace, creating a recursive loop.

### Workaround

The workaround was:

1. Press `N` when the main installer asks to install Marketplace.
2. Close and reopen PowerShell.
3. Confirm Spicetify is available:

```powershell
spicetify -v
```

4. Install Marketplace separately:

```powershell
iwr -useb https://raw.githubusercontent.com/spicetify/marketplace/main/resources/install.ps1 | iex
```

After doing this, the installation worked correctly.

### Possible cause

This seems to be caused by the Marketplace installer checking for `spicetify` using the current session PATH immediately after the CLI installer modifies the user PATH. On Windows, the current PowerShell session may not immediately see the updated PATH.

The fallback logic then tries to install Spicetify again, but since the main installer also asks whether to install Marketplace, this creates a loop.

### Suggested fix

A possible fix would be to update the current process PATH after installing Spicetify, for example by adding the Spicetify install directory to `$env:Path` before launching the Marketplace installer.

Another possible fix would be for the Marketplace installer to check the default install path directly, such as:

```powershell
$env:LOCALAPPDATA\spicetify\spicetify.exe
```

before deciding that Spicetify is not installed.

It may also help to prevent the Marketplace installer from recursively launching the main installer when it was already called from the main installer.

### 📸 Screenshots

_No response_

Cycle1337:
## Summary
- keep the existing process `PATH` when adding the Spicetify install directory, so child scripts can resolve `spicetify` in the same PowerShell session
- remove the old install directory from the process `PATH` as well as the user `PATH`
- fix the `UseBasicParsing` splatted parameter typo in the Windows installer download step

Fixes #3871

## Tests
- parsed `install.ps1` with `System.Management.Automation.Language.Parser`
- verified `Invoke-WebRequest` exposes `UseBasicParsing` in Windows PowerShell 5.1
- `git -c core.whitespace=blank-at-eol,blank-at-eof,space-before-tab,cr-at-eol diff --check`
- `go test ./...`

Note: I did not run the full installer locally to avoid modifying the machine's Spicetify/Spotify installation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the install/download process so web requests are configured correctly, helping avoid installation issues.
  * Fixed PATH updates to reliably remove old entries and add the app location to both the current session and future sessions, making the app easier to run after setup.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -95,7 +95,7 @@ function Get-Spicetify {
     Write-Host -Object "Downloading spicetify v$targetVersion..." -NoNewline
     $Parameters = @{
       Uri            = "https://github.com/spicetify/cli/releases/download/v$targetVersion/spicetify-$targetVersion-windows-$architecture.zip"
-      UseBasicParsin = $true
+      UseBasicParsing = $true
       OutFile        = $archivePath
     }
     Invoke-WebRequest @Parameters
@@ -113,16 +113,22 @@ function Add-SpicetifyToPath {
     Write-Host -Object 'Making spicetify available in the PATH...' -NoNewline
     $user = [EnvironmentVariableTarget]::User
     $path = [Environment]::GetEnvironmentVariable('PATH', $user)
+    $processPath = $env:PATH
   }
   process {
     $path = $path -replace "$([regex]::Escape($spicetifyOldFolderPath))\\*;*", ''
     if ($path -notlike "*$spicetifyFolderPath*") {
       $path = "$path;$spicetifyFolderPath"
     }
+
+    $processPath = $processPath -replace "$([regex]::Escape($spicetifyOldFolderPath))\\*;*", ''
+    if ($processPath -notlike "*$spicetifyFolderPath*") {
+      $processPath = "$processPath;$spicetifyFolderPath"
+    }
   }
   end {
     [Environment]::SetEnvironmentVariable('PATH', $path, $user)
-    $env:PATH = $path
+    $env:PATH = $processPath
     Write-Success
   }
 }


rxri:
do the nitpick comment that coderabbit mentioned

rxri:
Implemented it in other way, also in bash script

PR #3883, opened 2026-07-04
fix(apply): support rspack bundle layout in spotify 1.2.93

RianMorningstar:
Spotify 1.2.93 switched the client bundle from webpack to rspack and restructured the files, which broke custom app injection and the wrapper API ("Cannot find symbol for Custom app React symbols" and a "Something went wrong" screen on launch).

What changed in 1.2.93:

- xpui.js is gone, the route table and React.lazy loaders now live in xpui-modules.js
- the chunk name and css enable maps live in the runtime file xpui-snapshot.js
- the chunk global was renamed from webpackChunkclient_web to rspackChunkclient_web

What this PR does:

- runs insertCustomApp on xpui-modules.js and injects each anchor in whichever file it appears, only the file holding the lazy loaders can raise a version mismatch error
- waits for either chunk global in spicetifyWrapper.js and pushes generated custom app chunks to whichever one exists, so older Spotify versions keep working
- uses the existing fnStr helper in waitForURI, some rspack module exports have no callable toString and crashed it
- retries ModifyFile writes and reports failures instead of silently dropping patches, Windows keeps a user-mapped section open on the bundle files briefly after Spotify exits

Tested on Windows 11 with Spotify 1.2.93.667: theme, extensions, custom apps (marketplace) and sidebar nav links all load and render. Known remaining gap: the Spicetify.URI.from/fromString/idToHex finders still don't match the new bundle internals, that needs its own follow-up.

## Summary by CodeRabbit

* **New Features**
  * Improved compatibility with newer Spotify clients by supporting both webpack and rspack runtime chunk formats when loading custom apps.

* **Bug Fixes**
  * Custom app router and related configuration injection is now more resilient to client bundle changes, reducing mismatch/false-failure scenarios.
  * Improved handling of URI helper detection/conversion to better survive client updates.
  * Reduced apply failures by retrying file writes before giving up, with clearer recovery guidance.

### DIFF
--- a/jsHelper/spicetifyWrapper.js
+++ b/jsHelper/spicetifyWrapper.js
@@ -519,12 +519,13 @@ const fnStr = (f) => {
 };
 
 (async function hotloadWebpackModules() {
-	while (!window?.webpackChunkclient_web) {
+	while (!(window?.webpackChunkclient_web || window?.rspackChunkclient_web)) {
 		await new Promise((r) => setTimeout(r, 50));
 	}
+	const chunkGlobal = window.webpackChunkclient_web || window.rspackChunkclient_web;
 
 	// Force all webpack modules to load
-	const require = webpackChunkclient_web.push([[Symbol()], {}, (re) => re]);
+	const require = chunkGlobal.push([[Symbol()], {}, (re) => re]);
 	while (!require.m) await new Promise((r) => setTimeout(r, 50));
 	console.log("[spicetifyWrapper] Waiting for required webpack modules to load");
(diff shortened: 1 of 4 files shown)


rxri:
I'm already working on it locally. Thanks for contribution but I won't accept it

afonsos971:
Hello, my name is Afonso. I'm having trouble using Spicetify, as three recurring errors appear when I try to run it. I've already updated the tool and tried reapplying the Spotify version, but the problem persists. Can you help me diagnose the cause?
image

PR #3887, opened 2026-07-06
fix(css): artist profile picture fix affecting all images

KamilWachnicki:
Follow up to #3885 

My previous fix affected all image wrappers which at first they didn't appear broken but when i visited the marketplace some extensions had broken shape, the fix was to revert the `border-radius:50%` to `6px `and wrap `.main-cardImage-imageWrapper` in a `where:()` letting Spotify's hashed CSS class that applies to only artists win over the one injected, now only the artists should see the circular card border and it will remain working when Spotify re-hashes it's classes.

Built and checked that the CSS wins over the injected one.

My bad for the previous PR.

## Summary by CodeRabbit

* **Style**
  * Updated main layout styling, including fixed-width grid calculations for improved consistency.
  * Refined player card image presentation: configurable border-radius for standard images (defaulting to rounded corners), while preserving fully circular rounding when the circular variant is used.
  * Adjusted image wrapper/shadow styling and improved metadata truncation for cleaner, single-line display.
  * Enhanced play button hover/focus interactions and corrected bullet marker rendering.


rxri:
These selectors are really bad in terms of performance so ig we should fix it in completely other way

KamilWachnicki:
I think i got what u were saying

KamilWachnicki:
also could i improve the preprocess file by using go:embed to split the css we append to another file, it's horrible trying to read that line

rxri:
See if any of the comments from coderabbit are applicable, and leave that file alone for now. Not worth refactoring

KamilWachnicki:
Replaced `content:"â€¢"` with a proper `"\2022"` escape character, cleaner border-radius handling imo, removed the extra parenthesis in width calculation, proper overflow handling. Built and tested with themes.

rxri (approved):

rxri:
thx

PR #3892, opened 2026-07-16
fix(utils): avoid panics when creating default ini sections

SkenSMasteR:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration setup resilience by reporting section-creation errors without terminating the application.
  * Continued processing available configuration sections when individual sections cannot be created.


rxri:
That's actually good that spicetify panics. Config shouldn't generate half-filled in

PR #3898, opened 2026-07-28
feat: added new css mappings for right sidebar

Adrien5902:
Spotify introduced a collapsing feature for the right sidebar with the now playing content info, so I added the mappings to the collapsed bar classes, the names appended to `Root__right-sidebar` where taken from official app js

## Summary by CodeRabbit

* **New Features**
  * Added support for right sidebar peek states, including expanded, collapsed, and deferred-collapsed views.
  * Added overlay content and controls for the right sidebar peek experience.


rxri:
We called right sidebar buddyFeed some time ago, so are you sure, any of it won't apply to the buddyFeed mappings?

Adrien5902:
I don't know maybe, but then that means those buddyFeed mappings may be outdated because they don't show up on my devtools, and I don't get any results other than the two buttons at the top next to the profile icon when searching for buddyFeed on the page

I'm referring to this bar :
image

rxri (approved):

rxri:
pls don't use past tense in commits next time...

PR #3903, opened 2026-08-06
docs(install): explain shell PATH activation

yassinsolim:
## Summary

- tell shell-installer users to open a new terminal or reload their shell profile before using `spicetify`
- print the profile path selected by the installer, including custom `ZDOTDIR` locations
- keep the manual-PATH fallback accurate for unsupported shells

## Why

The shell installer updates a startup file, but when invoked through `curl | sh` it cannot update the already-running parent shell. The existing completion message immediately suggests running `spicetify --help`, which results in `command not found` until the user opens a new terminal or reloads the profile.

Windows is unaffected: `install.ps1` updates `$env:PATH` in the current PowerShell process as well as the persistent user PATH.

## Verification

- `sh -n install.sh`
- controlled installer smoke tests for zsh with a custom `ZDOTDIR`, bash, and fish
- Marketplace subprocess smoke test confirming the installer's exported PATH still resolves `spicetify`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation completion guidance based on whether a shell profile is detected.
  * Added profile reload instructions when applicable.
  * Provided `PATH` setup instructions when no usable shell profile is available.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -146,7 +146,13 @@ esac
 
 echo
 log "spicetify v$tag was installed successfully to $spicetify_install"
-log "Run 'spicetify --help' to get started"
+if [ -n "${shellrc:-}" ] && [ -f "$shellrc" ]; then
+    log "Open a new terminal or reload your shell profile to use spicetify:"
+    log "  source \"$shellrc\""
+    log "Then run 'spicetify --help' to get started"
+else
+    log "After adding spicetify to your PATH, run 'spicetify --help' to get started"
+fi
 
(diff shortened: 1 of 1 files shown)


yassinsolim:
Hi @rxri, CodeRabbit suggested you as the reviewer for this small installer UX fix. When you have a chance, would you mind taking a look? Thanks!

rxri (approved):

Toji-20:
excuse me how to update to 2.45.0

Issue #3090, opened 2024-07-04
modals broken


ohitstom:
### 🔍 Have you checked Spicetify.app page for your issue?

- [X] I have checked the FAQ

### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### ℹ Environment / Computer Info

```Markdown
Spotify for Windows (64 bit)
1.2.41.434.g39a25e2c
Spicetify v2.36.14
Theme: Comfy / Comfy
Extensions: npvAmbience.js, toggleDJ.js, gamepad.js, spotifyBackup.js, playbarClock.js, quickQueue.js, scannables.js, sleepTimer.js, tracksToEdges.js, adblock.js, immersiveView.js, volumePercentage.js
Custom apps: lyrics-plus, new-releases, marketplace
xpui_2024-06-28_1719586782170_39a25e2
cef_125.0.19+g3d8f1c9+chromium-125.0.6422.112
Runtime: Chrome
Build System: CMake
```

### 📝 Description

Updated spotify and spice at the same time, modals are still broken

### 📸 Screenshots

![image](https://github.com/spicetify/cli/assets/22730962/ac26e17f-f02e-41b4-9c20-fdcbceebc5bb)

ohitstom:
![image](https://github.com/spicetify/cli/assets/22730962/6e9ec62a-ec91-48e5-bf77-3b983259a006)
![image](https://github.com/spicetify/cli/assets/22730962/f453365f-0d7c-41f2-a161-a4e830325e0a)

sanoojes:
am also facing same issues updated the css map and reinstalled spicetify and spotify is is fixed ?

![image](https://github.com/spicetify/cli/assets/93819264/2565d049-0855-483a-a670-afbc41c43638)

i added some custom css to fix it for now 

```css
generic-modal {
  height: 100vh;
  width: 100vw;
  position: fixed;
  display: flex;
  align-items: center;
  justify-content: center;
}

.GenericModal__overlay {
  display: flex;
  width: 100%;
  z-index: 100;
  height: 100%;
  justify-content: center;
  align-items: center;
}
```

PR #3917, opened 2026-09-04
fix(wrapper): make PopupModal styling self-contained

theRealPadster:
### Problem

`PopupModal` builds its markup from Spotify's old Track Credits modal. Spotify has since deleted that component, so most of the classes in the template carry no styling at all. Auditing every stylesheet in `xpui/` on **1.2.98.301**, only two of seven survive:

| Class | Styled on 1.2.98? |
|---|---|
| `.GenericModal__overlay` | ✅ |
| `.main-embedWidgetGenerator-container` | ✅ |
| `.GenericModal` | ❌ |
| `.main-trackCreditsModal-container` | ❌ |
| `.main-trackCreditsModal-header` | ❌ |
| `.main-trackCreditsModal-closeBtn` | ❌ |
| `.main-trackCreditsModal-mainSection` | ❌ |
| `.main-type-alto` | ❌ |

Every Spicetify modal — ours and every extension's — therefore renders with an unstyled shell: no padding, the close button falling into normal flow **below** the title wearing default `<button>` chrome, an oversized default `h1`, and no scroll container, so tall content runs off the bottom of the window with no way to reach it. `isLarge: false` modals fare worst; their container has no styling left whatsoever.

This is a repeat: #3090 ("modals broken") was the same class of breakage in 2024.

### Why not css-map

This was my first instinct, and it doesn't work:

- The map already carries **41** `main-trackCreditsModal-*` entries. They're inert — the component was *removed*, not renamed, so there's no hash left to rewrite. `main-type-alto` has zero entries.
- The map has `zogFp9G1AEqb8AKOd5B0 → GenericModal`, but 1.2.98's modal box is `w1gLwZlc_sWpQ3CJ`. Adding that hash looks like a one-line fix — except `.w1gLwZlc_sWpQ3CJ` carries exactly one declaration, `-webkit-app-region: no-drag`. It's the bare react-modal base class.

Modern Spotify styles each dialog's *own* container rather than a shared modal box; the only thing dialogs still share is `__overlay`. There is nothing generic left to map onto.

Borrowing another live dialog's classes (`main-playlistEditDetailsModal-*`) would work today, but that's the same bet that just lost — Track Credits was live once too. It resets the clock instead of stopping it.

### Fix

Own the styling, as we already do for `ScrollableContainer`. Layout moves onto `spicetify-popup-*` classes backed by an injected stylesheet, following the existing `custom-apps.js` / `scrollable-container.js` pattern.

Two properties worth calling out:

- **Themes keep working.** The legacy class names stay on the elements, and every rule is wrapped in `:where()` so it contributes no specificity. A theme's plain `.main-trackCreditsModal-header { … }` still wins, with no `!important` needed.
- **`isLarge: true` still defers to Spotify.** It keeps `main-embedWidgetGenerator-container`, which Spotify does still style; our matching rule sits underneath as a fallback for when that one goes too.

### Verification

Against Spotify **1.2.98.301** over CDP, A/B against a wrapper built from the same tree *without* this change, with Marketplace's stylesheet disabled so only the wrapper is under test:

| | Control | With fix |
|---|---|---|
| header `display` | `block` | `flex` |
| header `padding` | `0px` | `24px` |
| title / close `y` | `90` / `134` (stacked) | `101` / `96` (same row) |
| close button | `2px` border, 28×33 | `0px` border, 32×32 |
| `h1` font-size | `32px` | `24px` |
| main `overflow-y` | `visible` | `auto` |
| border-radius | `0px` | `8px` |

Also confirmed:
- a simulated theme rule (`padding: 4px`, `border-radius: 0`) overrides the defaults without `!important`
- `isLarge: false` is now styled — 524px, background, 8px radius, flex column
- `isLarge: true` still measures Spotify's 664px container

`pnpm lint` and `pnpm fmt --check` clean. Source-only change; `jsHelper/spicetifyWrapper.js` is gitignored and built at release time.

`v3-beta` carries an identical `popup.js`, so this cherry-picks cleanly if wanted there.

### Related

Marketplace has a stopgap for its own modals in spicetify/marketplace#1235. That only helps users who have Marketplace installed; this fixes it for everyone, and that one can be dropped once this ships.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ds6ryy1Edpg9MS5VcB6Dmc

## Summary by CodeRabbit

* **Style**
  * Improved popup modal layout and visual consistency.
  * Added clearer styling for overlays, dialogs, headers, titles, close controls, and content areas.
  * Applied popup styling when a modal is first displayed.
  * Preserved compatibility with Spotify’s existing themes and styling.
  * Improved handling of large and standard popup layouts.

### DIFF
--- a/src/jsHelper/spicetifyWrapper/popup.js
+++ b/src/jsHelper/spicetifyWrapper/popup.js
@@ -1,8 +1,5 @@
-let stylesInjected = false;
-
 function injectStyles() {
-  if (stylesInjected) return;
-  stylesInjected = true;
+  if (document.querySelector("style.spicetify-popup-modal")) return;
 
   const style = document.createElement("style");
   style.className = "spicetify-popup-modal";
--- a/src/jsHelper/spicetifyWrapper/popup.js
+++ b/src/jsHelper/spicetifyWrapper/popup.js
@@ -1,3 +1,30 @@
+let stylesInjected = false;
+
+// Spotify removed the Track Credits modal component its markup was borrowed from,
+// so the classes below no longer carry any styling from Spotify's stylesheets.
+// They are kept on the elements for theme compatibility, while the `spicetify-popup-*`
+// classes own the layout. Rules are wrapped in `:where()` so they contribute no
+// specificity -- themes and extensions override them without needing `!important`.
+function injectStyles() {
+  if (stylesInjected || document.querySelector("style.spicetify-popup-modal")) return;
+  stylesInjected = true;
+
+  const style = document.createElement("style");
+  style.className = "spicetify-popup-modal";
+  style.textContent = `
+generic-modal :where(.spicetify-popup) { display: flex; max-height: calc(100vh - 64px); border-radius: 8px; overflow: hidden; }
+generic-modal :where(.spicetify-popup-container) { display: flex; flex-direction: column; width: 524px; max-width: 100%; min-height: 0; border-radius: 8px; color: var(--spice-text, var(--text-base, #fff)); background-color: var(--spice-player, var(--background-elevated-base, #121212)); }
+generic-modal :where(.spicetify-popup-container-large) { width: 664px; }
+generic-modal :where(.spicetify-popup-header) { display: flex; flex: 0 0 auto; align-items: center; justify-content: space-between; gap: 16px; padding: 24px; }
+generic-modal :where(.spicetify-popup-title) { margin: 0; font-size: 1.5rem; font-weight: 700; line-height: 1.3; }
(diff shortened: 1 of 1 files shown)


src/jsHelper/spicetifyWrapper/popup.js
```diff
@@ -1,19 +1,52 @@
+let stylesInjected = false;
+
+// Spotify removed the Track Credits modal component its markup was borrowed from,
+// so the classes below no longer carry any styling from Spotify's stylesheets.
+// They are kept on the elements for theme compatibility, while the `spicetify-popup-*`
+// classes own the layout. Rules are wrapped in `:where()` so they contribute no
+// specificity -- themes and extensions override them without needing `!important`.
```
rxri:
```suggestion
```

src/jsHelper/spicetifyWrapper/popup.js
```diff
@@ -1,19 +1,52 @@
...
+function injectStyles() {
+  if (stylesInjected || document.querySelector("style.spicetify-popup-modal")) return;
+  stylesInjected = true;
```
rxri: both do the same thing

rxri (changes requested):

rxri (approved):

PR #3944, opened 2026-09-15
feat: add per-command help support (-h <command>)

stefaceriani:
PR based on: https://github.com/spicetify/cli/issues/3943

### What changes
Adds the ability to show help for a single command, instead of only the full general help.

**Before:**
```
spicetify -h apply
```
always displayed the full help text (`help()`), regardless of the command passed — the only special case was `config`.

**After:**
```
spicetify -h apply
```
displays a dedicated help text for the `apply` command only, including its `Usage:` section. The same applies to every other supported command.

### Implementation
- Introduced a `commandHelp map[string]string`, with one entry per command (`backup`, `apply`, `refresh`, `restore`, `clear`, `enable-devtools`, `watch`, `restart`, `spotify-updates`, `path`, `color`, `config-dir`, `upgrade`, `update`).
- Introduced the `printHelp(command string)` function, which:
  - if `command == "config"` → calls `helpConfig()` (unchanged behavior)
  - if `command` exists in `commandHelp` → prints the dedicated text
  - otherwise → falls back to `help()` (unchanged behavior for unknown commands or when no command is specified)
- The `-h`/`--help` block in `init()` now calls `printHelp(kind)` instead of the inline logic `if kind == "config" { helpConfig() } else { help() }`.

### Unchanged behavior
- `spicetify -h` (no command) → still shows the full general help.
- `spicetify -h config` → unchanged, shows `helpConfig()`.
- No changes to flag/command parsing logic, nor to chainable/non-chainable commands.

### Why
Improves CLI usability: users can quickly get the syntax and options for a single command without scrolling through the entire general help output.

### Notes for reviewer
- Additive, backward-compatible change: no existing behavior is affected for users who don't use `-h <command>`.
- Worth considering whether future new commands should be added to `commandHelp` as well — otherwise they'll simply fall back to `help()`.

## Summary by CodeRabbit

* **Documentation**
  * Improved the `-h`/`--help` experience with command-specific help text.
  * Preserved dedicated configuration help and general help fallback behavior for unrecognized or unspecified commands.


rxri:
There's no need for per command help. `spicetify -h config` is only for config, as said in the help printout:
> For config information, run "spicetify -h config".

